PR #18, opened 2021-03-09
Create geopoint interface

ravifrancesco:
Added the basic Point class as well as the POIPoint class which contains a constructor that takes a GeoPoint as input


bgabriel998 (approved):
All good !

magangloff (approved):

PR #19, opened 2021-03-09
Implemented Google Sign API

giommok:
I have modified the button 2 activity by implementing Google Sign API with basic graphics from Google (so I added some strings, colors and images, like Google logo).
Also, I have added a keystore in order to make the API work (through Google Developer Console). Because of this, I had to add signingConfigs inside the app build.gradle (just for testing purposes, otherwise it wouldn't work on all our devices). This must be modified later if we will upload this app on Google Play Store (as the release signing-certificate fingerprint needed for Google API credentials is not the regular SHA-1 signing keys).


giommok:
> Looks good! Should we have the "// [START ...] // [END ...] in every file?

Actually, it is just how Google commented the necessary code. I included all those comments for this first commit, but unless you think it is smart, I was planning to remove all these "unnecessary" comments and personalize the code more for the next commit before Friday. The idea behind this commit was to make sure that sign-in worked for everyone, so we can show it on Friday without problems.

ravifrancesco (approved):
All seems fine

PR #20, opened 2021-03-09
Get poi from api

bgabriel998:
# Pull request Iusse #7 + #3 Geonames API
Add GeonamesHandler class.
Add Permissions to Manifest (Internet).
Add org.osmdroid.bonuspack project to the repo. 

___
`GET_POI_MAX_RESULTS` : Number of results requested to the API (all POI categories: not only mountains. The value is fixed to 200. May need to be modified by the user ? 
`GET_POI_MAX_RANGE` : Set to 20km, same as previous -> May need to be modified by the user ? 

`getSurroundingPeaks` : method is creating a thread that is calling API. Should the `thread.start()` and `thread.join()` method calls be in separate functions ?


magangloff (approved):
All good for me. I think that the range in km should be choosable by the user but the max_results should be fixed.

PR #23, opened 2021-03-10
Refactor Getsurrounding peaks. add possibility to choose max result o…

bgabriel998:
# Branch get_POI_from_API Refactor 
- Add possibility to change range and max number of result for each query 
- Split the `getSurroundingPeaks` function in two functions. The first one start the querry with `thread.start`, the second one call `thread.join` and populate the result list. The result list is then returned


giommok (approved):
All good

aolsson711 (approved):
Looks good

PR #24, opened 2021-03-11
Integrated Firebase

giommok:
- Supported Firebase
- Updated sign-in with Firebase Auth
- Modified Account class (from `GoogleAccount` to `FirebaseAccount`) to make it more general and to use Firebase User methods instead of Google Account ones (even if Google is our only provider at the moment)
- Added google_services.json to connect our app with Firebase


aolsson711 (approved):
Look good

bgabriel998 (approved):
Checked

PR #25, opened 2021-03-11
Extract user location

ravifrancesco:
Added classes UserPoint, GPSTracker, and UserLocationHandler.


giommok (approved):
All good

code/app/src/main/java/com/github/ravifrancesco/softwaredevproject/GPSTracker.java
```diff
@@ -0,0 +1,193 @@
...
+    }
+
+    // returns a boolean that indicates if the permits for accessing location are ok
+    public boolean canGetLocation() {
+        return this.canGetLocation;
+    }
+
+    @Override
+    public void onLocationChanged(Location location) {
+    }
```
bgabriel998, line 174: Don't we want to use the onLocationChanged callback to update the lat.long.alt. fields automatically ?

bgabriel998 (changes requested):

code/app/src/main/java/com/github/ravifrancesco/softwaredevproject/GPSTracker.java
ravifrancesco: You're right! I'll try to do some modifications

bgabriel998 (approved):
Approved

ravifrancesco:
Outdated

PR #27, opened 2021-03-11
Extract user location

ravifrancesco:
Added classes UserPoint, GPSTracker.


ravifrancesco:
Outdate

PR #28, opened 2021-03-11
Extract user location

ravifrancesco:
Added classes UserPoint and GPSTracker.


bgabriel998 (approved):
Everything is fine

PR #32, opened 2021-03-12
Camera view

magangloff:
Added camera preview on button 1


code/app/src/main/java/com/github/bgabriel998/softwaredevproject/Button1Activity.java
```diff
@@ -1,14 +1,54 @@
...
+            } catch (Exception e) {
+                e.printStackTrace();
+            }
+        }, ContextCompat.getMainExecutor(this));
+    }
+
+
+    private void bindImageAnalysis(@NonNull ProcessCameraProvider cameraProvider) {
+        ImageAnalysis imageAnalysis =
+                new ImageAnalysis.Builder().setTargetResolution(new Size(1440, 1080))
```
aolsson711, line 45: Are theese magic numbers something we want to have everytime. Then maybe they should be constants. Or are they just temporary?

code/app/src/main/java/com/github/bgabriel998/softwaredevproject/Button1Activity.java
```diff
@@ -1,14 +1,54 @@
...
+        previewView = findViewById(R.id.view_finder);
+        cameraProviderFuture = ProcessCameraProvider.getInstance(this);
+        cameraProviderFuture.addListener(() -> {
+            try {
+                ProcessCameraProvider cameraProvider = cameraProviderFuture.get();
+                bindImageAnalysis(cameraProvider);
+            } catch (Exception e) {
+                e.printStackTrace();
+            }
+        }, ContextCompat.getMainExecutor(this));
```
aolsson711, line 39: Would be nice with some comment on what this part of the code does. I am a bit unsure on what to review since I don't really understand this.

code/app/src/main/java/com/github/bgabriel998/softwaredevproject/Button1Activity.java
```diff
@@ -1,14 +1,54 @@
...
+    }
+
+
+    private void bindImageAnalysis(@NonNull ProcessCameraProvider cameraProvider) {
```
aolsson711, line 43: This function would also benefit from some comments. Alot is happening but I don't know what.

aolsson711 (commented):
I think everything looks good however it is hard to judge. Maybe a few comments here and there wouldn't hurt.

ravifrancesco (approved):
All good

PR #49, opened 2021-03-15
Removed camera preview test

magangloff:
Moved code folder to root, updated cirrus.yml file and test files.


aolsson711 (approved):
Tests needs to fix but, forced push so we can split work

.idea/jarRepositories.xml
```diff
@@ -26,5 +26,10 @@
       <option name="name" value="maven" />
       <option name="url" value="https://jitpack.io/" />
     </remote-repository>
+    <remote-repository>
+      <option name="id" value="maven" />
+      <option name="name" value="maven" />
+      <option name="url" value="https://jitpack.io" />
+    </remote-repository>
```
bgabriel998, line 33: Duplicate here, is that normal ?

app/build.gradle
```diff
@@ -11,7 +11,7 @@ android {
         targetSdkVersion 30
         versionCode 1
         versionName "1.0"
-        testApplicationId "com.github.bgabriel998.softwaredevproject"
+        //testApplicationId "com.github.bgabriel998.softwaredevproject"
```
bgabriel998, line 14: I think the test application Id is needed to run instrumented tests

app/src/androidTest/java/com/github/bgabriel998/softwaredevproject/ExampleInstrumentedTest.java
```diff
@@ -29,19 +29,4 @@ public void useAppContext() {
...
-        ArrayList<POI> result = null;
-        try {
-            result = handler.getSurroundingPeaks(point);
-            assertThat(result.size(), greaterThan(10));
-        } catch (InterruptedException e) {
-            e.printStackTrace();
-        }
-    }
-
-
```
bgabriel998, line 46: Note: removed test because the method doesn't exist anymore

ravifrancesco (approved):
Nice job, work needed to pass the codeclimate tests

bgabriel998 (approved):
See testc overage. Otherwise everything is fine

Issue #36, opened 2021-03-12
#4 Connect the UI with Sign In Feature


aolsson711:
As of now, the UI is just placeholders. The goal is to connect each button with the features that were implemented last week. That way we will get rid of the extra buttons

PR #63, opened 2021-03-21
Connect UI with sign in

aolsson711:
Copied the sign in part from button2 into the profile activity. Also removed SignInActivity since we want the sign in part in the profile. 

cloe #36


giommok (commented):
Seems good to me. Should we remove the unused stuff in Cirrus CI warnings?

giommok (approved):

magangloff (approved):
Seems good to me, have the tests for the SigIn activity also have been changed to use the new sign-in activity instead of the button2 activity?

aolsson711:
> Seems good to me, have the tests for the SigIn activity also have been changed to use the new sign-in activity instead of the button2 activity?

There were no tests for the sign in activity. I think we should research how to test the authentication

PR #66, opened 2021-03-22
Creating gps tests

ravifrancesco:
Added default location in GPStracker (mt. Everest)
Added GPStracker and UserPoint androidTests


app/src/androidTest/java/com/github/ravifrancesco/softwaredevproject/UserPointTest.java
```diff
@@ -0,0 +1,87 @@
...
+
+        UserPoint userPoint = new UserPoint(mContext);
+
+        if (userPoint.canGetLocation()) {
+            return;
+        } else {
+            Assert.assertEquals(27.988056, userPoint.getLatitude(), 0);
+            Assert.assertEquals(86.925278, userPoint.getLongitude(), 0);
+            Assert.assertEquals(8848.86, userPoint.getAltitude(), 0);
+            Assert.assertEquals(0.0, userPoint.getAccuracy(), 0);
```
aolsson711, line 39: If the code has a tolerance to 0. Is that the same thing as only having two arguments?
```suggestion
            Assert.assertEquals(27.988056, userPoint.getLatitude());
            Assert.assertEquals(86.925278, userPoint.getLongitude());
            Assert.assertEquals(8848.86, userPoint.getAltitude());
            Assert.assertEquals(0.0, userPoint.getAccuracy());
```

app/src/androidTest/java/com/github/ravifrancesco/softwaredevproject/UserPointTest.java
```diff
@@ -0,0 +1,87 @@
...
+import java.util.concurrent.TimeoutException;
+
+
+public class UserPointTest {
+
+    @Rule
+    public final ServiceTestRule serviceRule = new ServiceTestRule();
+
+    @Test
+    public void getLocationTest() throws TimeoutException {
```
aolsson711, line 23: If I understand the test correclty it tests that the default value is working not the getLocation funciton. If that is the case then maybe change testname. 
```suggestion
    public void defaultLocationTest() throws TimeoutException {
```

app/src/androidTest/java/com/github/ravifrancesco/softwaredevproject/UserPointTest.java
```diff
@@ -0,0 +1,87 @@
...
+        Context mContext = ApplicationProvider.getApplicationContext();
+
+        UserPoint userPoint = new UserPoint(mContext);
+
+        Random r = new Random();
+
+        double lat = r.nextDouble();
+        double lon = r.nextDouble();
+        double alt = r.nextDouble();
+        double acc = r.nextDouble();
```
aolsson711, line 60: I see the point with trying random values all the time. But as the tests are already slow, maybe just setting a constant is better for performance. It shouldn't differ the test if we always try a constant compared to a value between 0 and 1.

app/src/androidTest/java/com/github/ravifrancesco/softwaredevproject/UserPointTest.java
```diff
@@ -0,0 +1,87 @@
...
+
+        if (!userPoint.canGetLocation()) {
+            Assert.assertEquals(27.988056, userPoint.getLatitude(), 0);
+            Assert.assertEquals(86.925278, userPoint.getLongitude(), 0);
+            Assert.assertEquals(8848.86, userPoint.getAltitude(), 0);
+            Assert.assertEquals(0.0, userPoint.getAccuracy(), 0);
+        }
```
aolsson711, line 67: I think that the test might be a bit weird with this. This test is supposed to test if the function can set it's location. But it will fail if there is something wrong with the default values. The default values are tested in an other test and shouldn't be needed to be tested again.

app/src/androidTest/java/com/github/ravifrancesco/softwaredevproject/UserPointTest.java
```diff
@@ -0,0 +1,87 @@
...
+        Assert.assertEquals(acc, userPoint.getAccuracy(), 0);
+
+        userPoint.switchToRealLocation();
+
+        if (!userPoint.canGetLocation()) {
+            Assert.assertEquals(27.988056, userPoint.getLatitude(), 0);
+            Assert.assertEquals(86.925278, userPoint.getLongitude(), 0);
+            Assert.assertEquals(8848.86, userPoint.getAltitude(), 0);
+            Assert.assertEquals(0.0, userPoint.getAccuracy(), 0);
+        }
```
aolsson711, line 83: Same comment here. That the test should only test one thing.

app/src/androidTest/java/com/github/ravifrancesco/softwaredevproject/UserPointTest.java
```diff
@@ -0,0 +1,87 @@
...
+
+        userPoint.setLocation(lat,lon,alt,acc);
+
+        Assert.assertEquals(lat, userPoint.getLatitude(), 0);
+        Assert.assertEquals(lon, userPoint.getLongitude(), 0);
+        Assert.assertEquals(alt, userPoint.getAltitude(), 0);
+        Assert.assertEquals(acc, userPoint.getAccuracy(), 0);
```
aolsson711, line 74:
```suggestion
        Assert.assertEquals(lat, userPoint.getLatitude());
        Assert.assertEquals(lon, userPoint.getLongitude());
        Assert.assertEquals(alt, userPoint.getAltitude());
        Assert.assertEquals(acc, userPoint.getAccuracy());
```

app/src/androidTest/java/com/github/ravifrancesco/softwaredevproject/UserPointTest.java
```diff
@@ -0,0 +1,87 @@
...
+
+        userPoint.switchToRealLocation();
+
+        if (!userPoint.canGetLocation()) {
+            Assert.assertEquals(27.988056, userPoint.getLatitude(), 0);
+            Assert.assertEquals(86.925278, userPoint.getLongitude(), 0);
+            Assert.assertEquals(8848.86, userPoint.getAltitude(), 0);
+            Assert.assertEquals(0.0, userPoint.getAccuracy(), 0);
+        }
+
```
aolsson711, line 84: Remove empty line

app/src/androidTest/java/com/github/ravifrancesco/softwaredevproject/UserPointTest.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+
```
aolsson711, line 86: Remove empty line

aolsson711 (commented):
Looks overall good. Some comments on the test

app/src/androidTest/java/com/github/ravifrancesco/softwaredevproject/UserPointTest.java
ravifrancesco: It checks if the GPStracker is able to get the location (if it is, I cannot check if the location is correct, so the test just passes). If it is not able to get the location the tests check if the default location is returned

app/src/androidTest/java/com/github/ravifrancesco/softwaredevproject/UserPointTest.java
ravifrancesco: If I understood correctly, this type of random values don't impact the performance at all

app/src/androidTest/java/com/github/ravifrancesco/softwaredevproject/UserPointTest.java
ravifrancesco: True. My idea was to check if the location is returned correctly at the beginning, then check if the setlocation sets the location correctly, then when the real location is re-enabled it checks if it is still able to get the real location. Does it make sense?

app/src/androidTest/java/com/github/ravifrancesco/softwaredevproject/UserPointTest.java
ravifrancesco: See comment above

app/src/androidTest/java/com/github/ravifrancesco/softwaredevproject/UserPointTest.java
ravifrancesco: Why?

app/src/androidTest/java/com/github/ravifrancesco/softwaredevproject/UserPointTest.java
ravifrancesco: Why?

app/src/androidTest/java/com/github/ravifrancesco/softwaredevproject/UserPointTest.java
aolsson711: I see what you mean. I can just see the issue with that the test could fail because of issues with the default values or that the "switchToRealLocation" function doesn't work. I might be a bit petty, but shouldn't the "switchToRealLocation" have a separate test then instead?

app/src/androidTest/java/com/github/ravifrancesco/softwaredevproject/UserPointTest.java
aolsson711: To make the code fewer lines. Don't see the point in having
"
    }

}
"
instad of 
"
    }
}
"

aolsson711 (commented):
Discussion regarding test

app/src/androidTest/java/com/github/ravifrancesco/softwaredevproject/UserPointTest.java
ravifrancesco: I think it makes the code more readable. Empty lines are not counted in methods length

app/src/androidTest/java/com/github/ravifrancesco/softwaredevproject/UserPointTest.java
ravifrancesco: The fact is that setLocation and switchToRealLocation are strongly connected

app/src/androidTest/java/com/github/ravifrancesco/softwaredevproject/UserPointTest.java
aolsson711: I get empty lines help in methods between statements etc. But not an empty line between the end brackets right?

app/src/main/java/com/github/ravifrancesco/softwaredevproject/UserPoint.java
```diff
@@ -26,20 +27,27 @@
  */
 public class UserPoint extends Point {
 
+
+    private final GPSTracker gpsTracker;
```
magangloff, line 31: gpsTracker is defined multiple times

app/src/androidTest/java/com/github/ravifrancesco/softwaredevproject/GPSTrackerTest.java
```diff
@@ -0,0 +1,53 @@
...
+
+
+@RunWith(AndroidJUnit4.class)
+@MediumTest
+public class GPSTrackerTest {
+
+    @Rule
+    public final ServiceTestRule serviceRule = new ServiceTestRule();
+
+    @Test
```
magangloff, line 28: Add some quick comments on what is exactly tested.

magangloff (approved):
It would not be bad to have some more documentation for the tests.

app/src/androidTest/java/com/github/ravifrancesco/softwaredevproject/UserPointTest.java
magangloff, line 23: What is the difference between getLocationTest and testGPSTracker? Both tests depend on the userPoint right?
Also add some comments for the tests

ravifrancesco: It just checks that the gpstracker passes the location correctly to the userPoint

app/src/androidTest/java/com/github/ravifrancesco/softwaredevproject/GPSTrackerTest.java
ravifrancesco: Done

aolsson711 (approved):
Look good, little empty line here and there but ok ;)

PR #67, opened 2021-03-23
Firebase realtime database - Small hotfix

giommok:
The removed exception was causing problems since it is impossible to handle it from the main thread as the onCancelled method is run on a different thread.


app/src/main/java/com/github/giommok/softwaredevproject/Database.java
```diff
@@ -64,11 +64,8 @@ public void onDataChange(DataSnapshot snapshot) {
 
             @Override
             public void onCancelled(@NonNull DatabaseError error) {
-                throw new DatabaseException();
             }
```
bgabriel998, line 67: Handle what needs to be done in the onCancelled callback before merging to master

magangloff (approved):

app/src/main/java/com/github/giommok/softwaredevproject/Database.java
giommok: Actually, there is nothing to handle: onCancelled is triggered when the permission to read/write data is denied and on the official tutorial they suggested putting a simple log instruction. As we don't need that, I would prefer to have nothing (but I can't remove it as the listener requires the onCancelled method).

bgabriel998 (approved):

PR #71, opened 2021-03-25
UI dynamic rankings

aolsson711:
Added that the ranking items loads dynamically with a temp list to later be part of the database


ravifrancesco (commented):
Looks good. What is that TODO?

aolsson711:
> Looks good. What is that TODO?

That when the database is part of the list we will get the real username not just "username" as it is right now

bgabriel998 (commented):
Very good, nothing to say

bgabriel998 (approved):

ravifrancesco (approved):

PR #84, opened 2021-03-27
Replaced strings, added username validation and tests for username validation and sign-in

giommok:
- Added username validation
- Added tests for username validation
- Added tests for sign-in
- Replaced all the strings of profile activity with strings in strings.xml


app/src/main/java/com/github/giommok/softwaredevproject/Account.java
```diff
@@ -52,4 +51,14 @@ public static Account getAccount() {
      * This method updates the username of the current account
      */
     public void synchronizeUsername();
+
+    /**
+     * Check if the username chosen by the user is valid.
+     * @param username inserted by the user.
+     * @return true if and only if the username is valid.
+     */
+    public static Boolean isValid(String username) {
```
ravifrancesco, line 60: if(username == null) return false;
        return username.matches("\\w{3,15}");

Could be written as
return (username == null) ? false : username.matches("\\w{3,15}");

ravifrancesco (commented):
Looks good, left a comment on a method. Also, in class ProfileActivityTest.java methods start with an upper case letter, I think it should be changed to lower case.

app/src/main/java/com/github/giommok/softwaredevproject/Account.java
giommok: Good point, it can be written in just one line. However, I will write it like this:
`return username != null && username.matches("\\w{3,15}");`
I've also changed all the methods name in ProfileActivityTest.

ravifrancesco: Yes nice!

ravifrancesco (approved):

magangloff (approved):
All good.

PR #86, opened 2021-04-01
Implemented rankings-accounts integration

giommok:
- Integrated rankings with actual data from DB
- Modified rankings tests with real data
- Changed access modifiers of `RankingItem` attributes to private

Also fixed `ProfileActivityTest`, `DatabaseTest` and `AccountTest`  for increased stability


app/src/androidTest/java/com/github/bgabriel998/softwaredevproject/RankingsActivityTest.java
```diff
@@ -50,57 +63,63 @@ public void TestToolbarBackButton(){
         assertSame(testRule.getScenario().getResult().getResultCode(), Activity.RESULT_CANCELED);
     }
 
-    /* Test that all elements in list view are at correct place and contains correct data */
+    /* Test that a mock user in list view is at correct place and contains correct data */
     @Test
-    public void TestContentOfListView(){
-        DataInteraction interaction =  onData(instanceOf(RankingItem.class));
+    public void TestContentOfListView() throws InterruptedException {
```
aolsson711, line 68: Regarding this test, the reason we had as many as 20 users is that then Espresso also tests that you are able to scroll the list view. 
Should we add more mock users to still test this feature. I don't think it will impact our test coverage and I am not sure if we are supposed to test those types of features. But it couldn't hurt. We can discuss this on discord.

aolsson711 (commented):
I think all looks good code-wise. I've added a question to discuss on discored before approving review.

aolsson711 (approved):
Code looks good and my comment is fixed

bgabriel998 (approved):
Approved!

PR #90, opened 2021-04-04
Get 3d map

ravifrancesco:
Implemented the elevation map for the line of sight.
The map is downloaded as .AAIGrid and then converted to a matrix of integers (each integer represents an elevation in meters.
Added documentation for the ElevationMap class. There are some //TODO, I will do them as a new sprint task.
The way updates are managed has to be decided once the implementation for the real camera activity is implemented


app/src/androidTest/java/com/github/ravifrancesco/softwaredevproject/ElevationMapTest.java
```diff
@@ -0,0 +1,72 @@
...
+
+import android.content.Context;
+
+import androidx.test.core.app.ApplicationProvider;
+
+import org.junit.Assert;
+import org.junit.Test;
+
+public class ElevationMapTest {
+
```
bgabriel998, line 11:
```suggestion
/**
 * Test geoTIFF API helper class constructor
 */
```

app/src/androidTest/java/com/github/ravifrancesco/softwaredevproject/ElevationMapTest.java
```diff
@@ -0,0 +1,72 @@
...
+        UserPoint userPoint = new UserPoint(mContext);
+
+        // testing if the constructor works correctly
+        ElevationMap elevationMap = new ElevationMap(userPoint);
+
+        Assert.assertTrue(true);
+
+    }
+
```
bgabriel998, line 25:
```suggestion
/**
 * Check that getTopography method does effectively return a non-null value
 */
```

app/src/androidTest/java/com/github/ravifrancesco/softwaredevproject/ElevationMapTest.java
```diff
@@ -0,0 +1,72 @@
...
+        userPoint.setLocation(GPSTracker.DEFAULT_LAT, GPSTracker.DEFAULT_LON,GPSTracker.DEFAULT_ALT, GPSTracker.DEFAULT_ACC);
+
+        ElevationMap elevationMap = new ElevationMap(userPoint);
+
+        // checking if there are no errors retrieving the AAIGrid
+        int[][] topographyMap = elevationMap.getTopographyMap();
+        Assert.assertFalse(topographyMap == null);
+
+    }
+
```
bgabriel998, line 42:
```suggestion
/**
 *  Check the changing of the location to get Topography Map
 * Check with coordinates under the treshold and above
 */
```

bgabriel998 (approved):
Approved, just add the suggested comments for the test file (if they are correct)

app/src/main/java/com/github/ravifrancesco/softwaredevproject/ElevationMap.java
```diff
@@ -0,0 +1,200 @@
...
+     * @param nCol          width of the matrix.
+     * @param responseObj   Scanner to read the HTTPResponse.
+     */
+    private void buildMapGrid(int nRow, int nCol, Scanner responseObj) {
+
+        // skip lines to get to first row
+        responseObj.nextLine();
+        responseObj.nextLine();
+        responseObj.nextLine();
+        responseObj.nextLine();
```
aolsson711, line 112:
```suggestion
        // skip lines to get to first row
        int linesToSkip = 4;
        for (int i =0; i < linesToSkip; i++) {
            responseObj.nextLine();
        }
```

app/src/main/java/com/github/ravifrancesco/softwaredevproject/ElevationMap.java
```diff
@@ -0,0 +1,200 @@
...
+     * This method handles the updating of the elevation map. First the distance between the center
+     * of the current bounding box and the current user location is computed.
+     *
+     * If less than MINIMUM_DISTANCE_FOR_UPDATE no new map is downloaded, otherwise a new map is
+     * downloaded.
+     */
+    private void updateBoundingBox() {
+
+        Log.d("3D MAP",  "Updating");
+        POIPoint oldBoundingCenter = new POIPoint(boundingBox.getCenterWithDateLine());
```
aolsson711, line 148: I am a bit unsure on how often this method is called. I therefore wonder if it would save a bit of computing if we create the boundingCenter and save it when when we set the bounding box. So we only need to create an POIPoint one time instead of every time updateBoundingBox is called.

app/src/androidTest/java/com/github/ravifrancesco/softwaredevproject/ElevationMapTest.java
```diff
@@ -0,0 +1,79 @@
...
+        int[][] topographyMap = elevationMap.getTopographyMap();
+        Assert.assertFalse(topographyMap == null);
+
+    }
+/**
+ *  Check the changing of the location to get Topography Map
+ * Check with coordinates under the treshold and above
+ */
+    @Test
+    public void updateMapTest() {
```
aolsson711, line 51: We have not really talked about how we want to do tests. So maybe it is just my personal opinion, we can discuss it on Discord ;) 
But I personally prefer if this is two tests. One for when we change the location to under the threshold and one where we change it above. That way if it is just one of the parts that has been broken only one test would fail and we would now what happened. 
But as I said maybe that is my personal opinion and not a must following convention.

aolsson711 (commented):
Over all a code looking code. Nice work. Some minor comments on one test, a small for loop and a placement of a calculation only.

app/src/main/java/com/github/ravifrancesco/softwaredevproject/ElevationMap.java
ravifrancesco: Good point, I'll change it

app/src/androidTest/java/com/github/ravifrancesco/softwaredevproject/ElevationMapTest.java
ravifrancesco: I think these tests should stay together as they test the same method (and the same if statement, with two different outcomes)

aolsson711 (approved):
All looks good, discussed test over discorded and decide to have it like this.

PR #104, opened 2021-04-08
Add peak to db

### LINKED ISSUE #10: As a user, I want to be able to log in to the app. To save the mountains I have collected.

### LINKED ISSUE #60: #10 Add methods to save and retrieve data on visited mountains

bgabriel998:
# Save discovered peaks to firebase database

Add methods to save and retrieve the discovered peaks. 
When new peaks are scanned by the user, the list of previously scanned peaks is checked and only the new peaks are added to the database (and taken in account to compute the new user score).
To be more efficient, a **copy** of the user profile is saved dynamically. This avoid the application to request to often to much data to the DB. 
This copy is synchronized with the database using a callback method. 

A new functionality has been added to the user score class. When a user discovers a peak higher than (1000,2000,3000...m) for the first time, a bonus is given. 
The list of discovered range is also saved to the database. 

Close #10  
Close #60


app/src/main/java/com/github/giommok/softwaredevproject/Database.java
```diff
@@ -19,6 +19,21 @@
 }
 
 public class Database {
+
+    public static final String FOLDER = "/";
```
giommok, line 23: I'm not sure about this: instead of a folder string we can just add '/' to the following folder strings (in other words, we can just add '/' at the end of CHILD_USERS, for example)

app/src/main/java/com/github/giommok/softwaredevproject/Database.java
```diff
@@ -42,6 +57,37 @@ public static void setChild(String path, List<String> childKeys, List<Object> va
...
     }
 
+
+    /**
+     * Sets an array of values to the given child at the given path in the DB
+     * @param path The path to the fields to be modified
+     * @param child child key where to put the new values
+     * @param values list of values to add
+     */
+    public static void setChildObjectList(String path, String child,  ArrayList<Object> values){
```
giommok, line 67: I don't think it's necessary to have both path and child parameters, as the child is part of the path.

app/src/main/java/com/github/giommok/softwaredevproject/Database.java
```diff
@@ -42,6 +57,37 @@ public static void setChild(String path, List<String> childKeys, List<Object> va
...
+    }
+
+
+    /**
+     * Sets an array of values to the given child at the given path in the DB
+     * @param path The path to the fields to be modified
+     * @param child child key where to put the new values
+     * @param value object to add
+     */
+    public static void setChildObject(String path, String child,  Object value){
```
giommok, line 83: Same comment as the last one.

app/src/main/java/com/github/giommok/softwaredevproject/FirebaseAccount.java
```diff
@@ -5,48 +5,70 @@
...
 
 /* Singleton class containing the only possible account connected */
 public class FirebaseAccount implements Account {
 
-    private static DatabaseReference dbRefUsername = Database.refRoot.child("users/null/username");
-    private static ValueEventListener usernameListener = new ValueEventListener() {
+    private static String username = "null";
+    private static long score = 0;
+    private static FirebaseAccount account = null;
+    private static DatabaseReference dbRefUser = Database.refRoot.child("users/null/username");
```
giommok, line 35: As the reference is to the user, I think it is better to use just the user as child.
```suggestion
    private static DatabaseReference dbRefUser = Database.refRoot.child("users/null");
```

app/src/main/java/com/github/ravifrancesco/softwaredevproject/POIPoint.java
```diff
@@ -28,11 +33,42 @@ public POIPoint(GeoPoint geoPoint) {
...
+     * @return
+     */
+    @Override
+    public boolean equals(Object obj) {
+        if (obj == null) return false;
+        if (!(obj instanceof POIPoint))
+            return false;
+        if (obj == this)
+            return true;
+        return Objects.equals(this.getName(), ((POIPoint) obj).getName());
```
giommok, line 72:
```suggestion
        return getName().equals(((POIPoint)obj).getName());
```
As both are strings, maybe we can use the strings equals.

giommok (changes requested):
I left some comments, everything else seems good.

app/src/main/java/com/github/giommok/softwaredevproject/Database.java
bgabriel998: Sometimes the string "users/" was used, and some other only "users" was used. I don't know if we can simply supress the "/" or add it everywhere without trouble

giommok: It should be fine if we add it everywhere, but anyway you can add it and see if the tests are good.

bgabriel998: Ok i'll add the "/" to the `CHILD_USERS` string

app/src/main/java/com/github/giommok/softwaredevproject/Database.java
bgabriel998: Changes made

app/src/main/java/com/github/giommok/softwaredevproject/Database.java
bgabriel998: Changes made

app/src/main/java/com/github/giommok/softwaredevproject/FirebaseAccount.java
```diff
@@ -93,97 +115,160 @@ public String getUsername() {
...
+
+        for (Map.Entry<String, HashMap<String, String>> entry : entries.entrySet()) {
+            Object value = entry.getValue();
+            POIPoint peak = new POIPoint(((HashMap<String, String>) value).get(Database.CHILD_ATTRIBUTE_PEAK_NAME),
+                    ((HashMap<String, Double>) value).get(Database.CHILD_ATTRIBUTE_PEAK_LATITUDE),
+                    ((HashMap<String, Double>) value).get(Database.CHILD_ATTRIBUTE_PEAK_LONGITUDE),
+                    ((HashMap<String, Long>) value).get(Database.CHILD_ATTRIBUTE_PEAK_ALTITUDE));
+
+            //Check if the peak is already contain to avoid duplicate creation
+            if(!discoveredPeaks.contains(peak)) discoveredPeaks.add(peak);
```
magangloff, line 217: HashSet.add() already checks if the object is already contained thus no need to check.
```suggestion
            discoveredPeaks.add(peak);
```

app/src/main/java/com/github/giommok/softwaredevproject/FirebaseAccount.java
```diff
@@ -93,97 +115,160 @@ public String getUsername() {
...
+     * Get the list of discovered peak heights (ranges)
+     * The list is automatically appended to the discoveredHeights hashSet
+     * @param discoveredHeightsFromHashMap snapshot from the database userchild
+     */
+    private static void syncGetDiscoveredHeight(DataSnapshot discoveredHeightsFromHashMap){
+        HashMap<String, HashMap<String, String>> entries = (HashMap<String, HashMap<String, String>>) discoveredHeightsFromHashMap.getValue();
+        for (Map.Entry<String, HashMap<String, String>> entry : entries.entrySet()){
+            Object value = entry.getValue();
+            long retrievedVal = ((ArrayList<Long>) value).get(0);
+            if(!discoveredPeakHeights.contains(retrievedVal)) discoveredPeakHeights.add((int)retrievedVal);
```
magangloff, line 271: Same here for the .add()
```suggestion
             discoveredPeakHeights.add((int)retrievedVal);
```

magangloff (approved):
Small comment, else all good!

giommok (approved):

PR #116, opened 2021-04-15
Added loading circle

giommok:
After Google Sign-In, connection to Firebase Servers takes a few seconds, therefore I added a loading circle that shows up in the meanwhile.


aolsson711 (approved):
Looks Good

bgabriel998 (approved):

PR #131, opened 2021-04-20
Gallery file path

aolsson711:
- Connected Gallery with file paths where photos taken with app is saved
- Improved toolbar back tests

### DIFF
--- a/app/src/androidTest/java/com/github/bgabriel998/softwaredevproject/FullGalleryActivityTest.java
+++ b/app/src/androidTest/java/com/github/bgabriel998/softwaredevproject/FullGalleryActivityTest.java
@@ -75,6 +75,6 @@ public class FullGalleryActivityTest {
                 .perform(RecyclerViewActions.actionOnItemAtPosition(19,
                         click()));
         intended(allOf(IntentMatchers.hasComponent(ImageActivity.class.getName()),
-                IntentMatchers.hasExtra("imagePath", path)));
+                IntentMatchers.hasExtra(ImageActivity.IMAGE_PATH_INTENT, path)));
     }
 }
--- a/app/src/androidTest/java/com/github/bgabriel998/softwaredevproject/FullGalleryActivityTest.java
+++ b/app/src/androidTest/java/com/github/bgabriel998/softwaredevproject/FullGalleryActivityTest.java
@@ -1,40 +1,52 @@
 package com.github.bgabriel998.softwaredevproject;
 
-import android.app.Activity;
-
-import androidx.test.espresso.DataInteraction;
-import androidx.test.espresso.Espresso;
-import androidx.test.espresso.ViewInteraction;
+import androidx.test.core.app.ApplicationProvider;
 import androidx.test.espresso.contrib.RecyclerViewActions;
 import androidx.test.espresso.intent.Intents;
 import androidx.test.espresso.intent.matcher.IntentMatchers;
 import androidx.test.espresso.matcher.ViewMatchers;
 import androidx.test.ext.junit.rules.ActivityScenarioRule;
+import androidx.test.ext.junit.runners.AndroidJUnit4;
 
 import org.junit.After;
+import org.junit.AfterClass;
 import org.junit.Before;
+import org.junit.BeforeClass;
 import org.junit.Rule;
 import org.junit.Test;
+import org.junit.runner.RunWith;
 
-import java.util.Locale;
-
-import static androidx.test.espresso.Espresso.onData;
 import static androidx.test.espresso.Espresso.onView;
 import static androidx.test.espresso.action.ViewActions.click;
 import static androidx.test.espresso.assertion.ViewAssertions.matches;
 import static androidx.test.espresso.intent.Intents.intended;
-import static androidx.test.espresso.matcher.ViewMatchers.hasDescendant;
+import static androidx.test.espresso.matcher.ViewMatchers.withEffectiveVisibility;
 import static androidx.test.espresso.matcher.ViewMatchers.withId;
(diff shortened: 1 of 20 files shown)


app/src/androidTest/java/com/github/bgabriel998/softwaredevproject/UITestHelper.java
```diff
@@ -99,4 +130,42 @@ private static Bitmap getBitmap(Drawable drawable){
...
+        try {
+            Drawable image = ContextCompat.getDrawable(ApplicationProvider.getApplicationContext(),
+                    R.drawable.temp_diablerets);
+            assert image != null;
+            Bitmap bitmap = getBitmap(image);
+
+            File imageFile = new File(CameraActivity.getOutputDirectory(ApplicationProvider.getApplicationContext()),
+                    filename);
+            FileOutputStream outputStream = new FileOutputStream(imageFile);
+            int quality = 100;
```
giommok, line 148: What if we make this parameter a final static attribute of the class `UITestHelper`? It would be easier to find and modify it, if necessary.

app/src/main/java/com/github/bgabriel998/softwaredevproject/GalleryActivity.java
```diff
@@ -27,35 +29,37 @@ protected void onCreate(Bundle savedInstanceState) {
...
+        List<String> imagePaths = getImagePaths();
+        if (!imagePaths.isEmpty()) {
+            findViewById(R.id.gallery_empty).setVisibility(View.GONE);
+        }
 
-        GalleryAdapter galleryAdapter = new GalleryAdapter(this, images, image -> {
+        GalleryAdapter galleryAdapter = new GalleryAdapter(this, imagePaths, imagePath -> {
             Intent intent = new Intent(this, ImageActivity.class);
-            intent.putExtra("image", image);
+            intent.putExtra("imagePath", imagePath);
```
giommok, line 41: For the same reason as the last comment, would it be convenient to make "imagePath" a final static attribute? This way, we could call the same attribute when the extra of the intent is used instead of writing the same string two times

app/src/main/java/com/github/bgabriel998/softwaredevproject/ImageActivity.java
```diff
@@ -18,16 +20,17 @@ protected void onCreate(Bundle savedInstanceState) {
         ToolbarHandler.SetupToolbar(this, TOOLBAR_TITLE);
 
         Intent intent = getIntent();
-        setImage(intent.getIntExtra("image", -1));
+        setImage(intent.getStringExtra("imagePath"));
```
giommok, line 23: This is where we could call the static attribute suggested in the last comment.

giommok (changes requested):
Nice code and well done with the Toolbar tests trick! 
I suggested a few minor changes in the comments.

magangloff (changes requested):
Scrolling through the gallery results in Jank (slow rendering).
Maybe it's worth checking https://developer.android.com/topic/performance/vitals/render#common-jank to see if you can identify the problem.
Added some request for the pictures that are overdrawn by the toolbar and the pictures that were displayed in the wrong orientation.

aolsson711 (commented):
See comments

app/src/androidTest/java/com/github/bgabriel998/softwaredevproject/UITestHelper.java
giommok: In this case, I think it's fine if we keep it as it is.

app/src/androidTest/java/com/github/bgabriel998/softwaredevproject/ImageActivityTest.java
```diff
@@ -7,26 +7,49 @@
...
-    private static final int testImage = R.drawable.temp_camera;
+    private static final int testImage = R.drawable.temp_diablerets;
+    private static final String imagePath = CameraActivity.getOutputDirectory(ApplicationProvider.getApplicationContext()) +
+                                                "/TestImage";
 
     static Intent intent;
     static {
         intent = new Intent(ApplicationProvider.getApplicationContext(), ImageActivity.class);
-        intent.putExtra("image", testImage);
+        intent.putExtra("imagePath", imagePath);
```
giommok, line 40:
```suggestion
        intent.putExtra(ImageActivity.IMAGE_PATH_INTENT, imagePath);
```
Discussion on discord!

giommok (changes requested):

app/src/androidTest/java/com/github/bgabriel998/softwaredevproject/FullGalleryActivityTest.java
```diff
@@ -48,29 +60,21 @@ public void cleanUp(){
...
     public void TestPressLastItem(){
-        // TODO Redo test when actually using database.
+        String path = CameraActivity.getOutputDirectory(ApplicationProvider.getApplicationContext()) +
+                      "/TestImage19";
         onView(withId(R.id.gallery_recyclerview))
                 .perform(RecyclerViewActions.actionOnItemAtPosition(19,
                         click()));
         intended(allOf(IntentMatchers.hasComponent(ImageActivity.class.getName()),
-                IntentMatchers.hasExtra("image", R.drawable.temp_camera)));
+                IntentMatchers.hasExtra("imagePath", path)));
```
giommok, line 78:
```suggestion
                IntentMatchers.hasExtra(ImageActivity.IMAGE_PATH_INTENT, path)));
```
Found the string here too!

giommok (approved):

magangloff (approved):

PR #141, opened 2021-04-22
Display mountain info

magangloff:
Added the name and the height of the mountain above the marker in a 45° angle


app/src/main/java/com/github/bgabriel998/softwaredevproject/CameraUiView.java
```diff
@@ -363,10 +372,25 @@ private void drawMountainMarker(POIPoint poiPoint, Boolean isVisible, int actual
...
+
+        canvas.drawBitmap(mountainMarker, left, mountainMarkerPosition, null);
+
+        //Save status before Screen Rotation
+        canvas.save();
+        canvas.rotate(LABEL_ROTATION, left, mountainMarkerPosition);
+        canvas.drawText(poiPoint.getName() + " " + poiPoint.getAltitude() + "m",
+                left + mountainInfo.getTextSize() - OFFSET_MOUNTAIN_INFO,
+                mountainMarkerPosition + mountainInfo.getTextSize() + OFFSET_MOUNTAIN_INFO,
+                mountainInfo);
```
aolsson711, line 391: It looks good. I just wondered if it is possible to create some kind of tests for this to see if the correct test is set?

aolsson711 (commented):
Good looks good, However I had a question regarding if it is possible to write a test to tests the displayed info.

app/src/main/java/com/github/bgabriel998/softwaredevproject/CameraUiView.java
```diff
@@ -28,24 +26,30 @@
...
     //Font size of the text
     private int mainTextSize;
 
     //Max opacity for the paints
     private static final int MAX_ALPHA = 255;
 
+    //Rotation to be applied for the addition info of the mountains
+    private static final int LABEL_ROTATION = -45;
+    //Offset for the text to be above the marker
+    private static final float OFFSET_MOUNTAIN_INFO = 15;
```
giommok, line 45: Maybe as it's a final attribute we can put it as an int, as it is not assigned a float number.
```suggestion
    private static final int OFFSET_MOUNTAIN_INFO = 15;
```

giommok (changes requested):
Looks good to me, just one comment on a minor detail.

app/src/main/java/com/github/bgabriel998/softwaredevproject/CameraUiView.java
magangloff: Well the text is the name of the POIPoint so to test if the text is correct, we need to test the POIPoint.getName() method which is probably already done.

giommok (approved):

aolsson711 (approved):
All looks good

PR #163, opened 2021-04-24
Updated ProfileActivity and improved Firebase tests

giommok:
#161
- Implemented anonymous Firebase users for testing purposes.
- Updated Firebase tests (`ProfileActivityTests`, `DatabaseTests`, `AccountTests`, `FriendsActivityTest`) so that multiple tests running at the same time do not use the same users and therefore do not conflict, by using Firebase anonymous users.
- Removed some `ProfileActivity` UI methods that are now unnecessary as testing with anonymous Firebase users is much more flexible.

#162 
- Reworked `ProfileActivity` with `registerForActivityResult` methods instead of deprecated `startActivityForResult`.
- Moved "add friend" feature to `AddFriendActivity` from `ProfileActivity`.

Also moved the classes into the proper packages.


aolsson711 (commented):
The code looks great, good job. However I don't seem to find any tests regarding the new add friends activity. That show that the correct result code is returned etc. Am i missing something or hasn't it been added ?

app/src/androidTest/java/com/github/giommok/softwaredevproject/AccountTest.java
```diff
@@ -171,44 +252,43 @@ public void setAndGetDiscoveredCountryHighPoints()throws InterruptedException{
...
         DatabaseReference refAdd = Database.refRoot.child(Database.CHILD_USERS);
-        refAdd.child("null").child(Database.CHILD_COUNTRY_HIGH_POINT).push().setValue(newEntry);
+        refAdd.child(account.getId()).child(Database.CHILD_COUNTRY_HIGH_POINT).push().setValue(newEntry);
 
-        Thread.sleep(1000);
+        Thread.sleep(sleepTime);
         String s1 = account.getDiscoveredCountryHighPoint().get("France").toString();
         String s2 = newEntry.toString();
-        assertEquals(s2,s1);
+        assertEquals(s1, s2);
```
bgabriel998, line 270: AssertEquals takes the expected value as first parameter and the actual value as second parameter. 
This is important for debugging purposes if the test fails : the assert string needs to be correct. 
```suggestion
        assertEquals(s2, s1);
```

app/src/androidTest/java/com/github/giommok/softwaredevproject/DatabaseTest.java
```diff
@@ -3,69 +3,61 @@
...
+        Database.isPresent(Database.CHILD_USERS, Database.CHILD_USERNAME, user1, () -> assertTrue("Correct behavior", true),  Assert::fail);
+        Thread.sleep(sleepTime);
     }
 
     /* Test that setChild method works */
     @Test
     public void setChildTest() throws InterruptedException {
-        Database.isPresent("users", "username", "testing@Value", Assert::fail, () -> assertTrue("Correct behavior", true));
+        Database.isPresent(Database.CHILD_USERS, Database.CHILD_USERNAME, user1, Assert::fail, () -> assertTrue("Correct behavior", true));
         Thread.sleep(1500);
```
bgabriel998, line 50: Replace value by the newly defined constant
```suggestion
        Thread.sleep(SLEEP_TIME);
```

app/src/androidTest/java/com/github/giommok/softwaredevproject/DatabaseTest.java
```diff
@@ -3,69 +3,61 @@
...
-        Database.setChild("users/null", Arrays.asList("username"), Arrays.asList("testing@Value"));
+        Database.setChild(Database.CHILD_USERS + user1, Collections.singletonList(Database.CHILD_USERNAME), Collections.singletonList(user1));
         Thread.sleep(1500);
```
bgabriel998, line 52: Replace value by the newly defined constant
```suggestion
        Thread.sleep(SLEEP_TIME);
```

app/src/androidTest/java/com/github/giommok/softwaredevproject/DatabaseTest.java
```diff
@@ -3,69 +3,61 @@
...
-        Database.isPresent("users", "username", "testing@Value", () -> assertTrue("Correct behavior", true), Assert::fail);
+        Database.isPresent(Database.CHILD_USERS, Database.CHILD_USERNAME, user1, () -> assertTrue("Correct behavior", true), Assert::fail);
         Thread.sleep(1500);
```
bgabriel998, line 54:
```suggestion
        Thread.sleep(SLEEP_TIME);
```

app/src/androidTest/java/com/github/giommok/softwaredevproject/DatabaseTest.java
```diff
@@ -3,69 +3,61 @@
...
     }
 
     /* Test that exception is thrown if different list sizes are provided */
-    @Test(expected = DatabaseException.class)
+    @Test(expected = RuntimeException.class)
     public void differentSizes_SetChildTest() throws InterruptedException {
-        Database.setChild("users/null", Arrays.asList("username", "error"), Arrays.asList("testingValue"));
+        Database.setChild(Database.CHILD_USERS + user1, Arrays.asList("string1", "string2"), Collections.singletonList(""));
         Thread.sleep(1000);
```
bgabriel998, line 61: Consider adding another constant (=1000)

bgabriel998 (changes requested):
Code is fine. I changed the constants names (in lower case) for upper case. That way we can differentiate constants from variables. It's a good thing that you added a constant for the thread sleep in the test file. This should become a standart for all test file.

**PLEASE CHECK THAT ALL CONSTANTS HAVE BEEN REFACTORED CORRECTLY**

_Also note that assertEquals takes the expected value as 1st parameter and actual value as 2nd parameter_

giommok:
> The code looks great, good job. However I don't seem to find any tests regarding the new add friends activity. That show that the correct result code is returned etc. Am i missing something or hasn't it been added ?

I left the `AddFriendActivity` tests in the `ProfileActivityTest` class without making a separate test class too, as the add friend intent is started with a `registerForActivityResult` from `ProfileActivity` and so I also tested the correct behaviour on `ProfileActivity` after the result is returned. 
Do you think it is okay or should I move also the tests to a new class?

giommok:
> Code is fine. I changed the constants names (in lower case) for upper case. That way we can differentiate constants from variables. It's a good thing that you added a constant for the thread sleep in the test file. This should become a standart for all test file.

You are totally right about the upper case! I will refactor the code directly from Android Studio.

> _Also note that assertEquals takes the expected value as 1st parameter and actual value as 2nd parameter_

Yes, indeed, I thought that s1 was the expected value and s2 the actual one. Sorry for that!

app/src/androidTest/java/com/github/giommok/softwaredevproject/AccountTest.java
giommok: Thanks for the correction!

app/src/androidTest/java/com/github/giommok/softwaredevproject/DatabaseTest.java
giommok: I removed this sleep as there was no need.

bgabriel998 (approved):

aolsson711 (approved):
Nice work! Checked code before Moving files. It looked good then

PR #164, opened 2021-04-26
Created fullscreen

### LINKED ISSUE #142: #92 Fix so all images are full screen when pressing them in the gallery
Currently there is some issue that some images aren't fullscreen but they should be. Maybe some scaling needs to be done

aolsson711:
Reworked ImageActivity to always be full screen. 
Removed toolbar and "small top bar" to make sure we see full image all the time. 

closes #142

### DIFF
--- a/app/src/main/java/com/github/bgabriel998/softwaredevproject/gallery/ImageActivity.java
+++ b/app/src/main/java/com/github/bgabriel998/softwaredevproject/gallery/ImageActivity.java
@@ -8,6 +8,7 @@ import android.graphics.BitmapFactory;
 import android.graphics.Matrix;
 import android.media.ExifInterface;
 import android.os.Bundle;
+import android.view.WindowManager;
 import android.widget.ImageView;
 
 import com.github.bgabriel998.softwaredevproject.R;
@@ -18,7 +19,6 @@ import java.io.IOException;
 public class ImageActivity extends AppCompatActivity {
 
     public static final String IMAGE_PATH_INTENT = "imagePath";
-    private static final String TOOLBAR_TITLE = "Image";
     private static final int ROTATE_DEGREES = 90;
 
     @Override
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/github/bgabriel998/softwaredevproject/gallery/ImageActivity.java
```diff
@@ -8,6 +8,7 @@
 import android.graphics.Matrix;
 import android.media.ExifInterface;
```
giommok, line 9: Cirrus suggests that we avoid `android.media.ExifInterface` and we replace it with the one I wrote in the suggestion.
```suggestion
import android.support.media.ExifInterface;
```

I found [this](https://stackoverflow.com/questions/44246110/android-error-while-using-exifinterface) as an explanation of why it is better. It may be necessary to add the following line in build.gradle, as stated in the linked resource:
`compile "com.android.support:exifinterface:25.1.0"`

giommok (changes requested):
Good job! It looks fine to me. 
I made a comment about a Cirrus warning that I noticed, let me know what you think!

app/src/main/java/com/github/bgabriel998/softwaredevproject/gallery/ImageActivity.java
aolsson711: According to your link it is an issue for Android versions lower than 24. We have min at 24. But I can still change it if you want ofcourse!

giommok: Oh you're right. I don't think it is necessary then.

giommok (approved):

ravifrancesco (approved):
Everything seems good to me. Codeclimate Is complaining but the page shows no issues

PR #165, opened 2021-04-27
Fixed bug markers disappearing near 0°

magangloff:
Fixed bug #153 
Mountain markers west of north (below 360) or now displayed when looking east of north (above 0°). Same goes other way around.

### DIFF
--- a/app/src/main/java/com/github/bgabriel998/softwaredevproject/camera/CameraUiView.java
+++ b/app/src/main/java/com/github/bgabriel998/softwaredevproject/camera/CameraUiView.java
@@ -342,7 +342,7 @@ public class CameraUiView extends View {
     private void drawPOIs(int actualDegree){
         //Go through all POIPoints
         for(POIPoint poiPoint : POIPoints){
-            if((int)poiPoint.getHorizontalBearing() == actualDegree){
+            if((int)poiPoint.getHorizontalBearing() == (actualDegree + 360) % 360){
                 drawMountainMarker(poiPoint, false, actualDegree);
             }
         }
@@ -356,7 +356,7 @@ public class CameraUiView extends View {
     private void drawLabeledPOIs(int actualDegree){
         //Go through all POIPoints
         labeledPOIPoints.entrySet().stream()
-                .filter(p -> (int)p.getKey().getHorizontalBearing() == actualDegree)
+                .filter(p -> (int)p.getKey().getHorizontalBearing() == (actualDegree + 360) % 360)
                         .forEach(p -> drawMountainMarker(p.getKey(), p.getValue(), actualDegree));
     }
 


app/src/main/java/com/github/bgabriel998/softwaredevproject/camera/CameraUiView.java
```diff
@@ -342,7 +342,7 @@ else if(POIPoints != null && !POIPoints.isEmpty()){
     private void drawPOIs(int actualDegree){
         //Go through all POIPoints
         for(POIPoint poiPoint : POIPoints){
-            if((int)poiPoint.getHorizontalBearing() == actualDegree){
+            if((int)poiPoint.getHorizontalBearing() == (actualDegree + 360) % 360){
```
aolsson711, line 345:
```suggestion
            if((int)poiPoint.getHorizontalBearing() == actualDegree % 360){
```

app/src/main/java/com/github/bgabriel998/softwaredevproject/camera/CameraUiView.java
```diff
@@ -356,7 +356,7 @@ private void drawPOIs(int actualDegree){
     private void drawLabeledPOIs(int actualDegree){
         //Go through all POIPoints
         labeledPOIPoints.entrySet().stream()
-                .filter(p -> (int)p.getKey().getHorizontalBearing() == actualDegree)
+                .filter(p -> (int)p.getKey().getHorizontalBearing() == (actualDegree + 360) % 360)
```
aolsson711, line 359:
```suggestion
                .filter(p -> (int)p.getKey().getHorizontalBearing() == actualDegree % 360)
```

aolsson711 (commented):
Remove + from modolo

app/src/main/java/com/github/bgabriel998/softwaredevproject/camera/CameraUiView.java
magangloff: I just checked online and in code and actually the modulo operator % in java returns negative values

aolsson711 (approved):
Looks good!

app/src/main/java/com/github/bgabriel998/softwaredevproject/camera/CameraUiView.java
giommok, line 345: I have a doubt about the `+ 360` part. I understand that % can return a negative value if actualDegree < 0, so we need it. 
But if an angle can be negative, are we sure that it is still always greater than -360? Because if `actualDegree` were less than -360, it would return a negative number even after adding 360 to it.
What if instead of adding 360 we do something like this?
```suggestion
            if((int)poiPoint.getHorizontalBearing() == Math.abs(actualDegree) % 360){
```

giommok (changes requested):
I have a question related to what @aolsson711 pointed out, let me know what you think.

app/src/main/java/com/github/bgabriel998/softwaredevproject/camera/CameraUiView.java
magangloff: The name of ```actualDegree``` is probably a bit confusing. It is not the heading of the user but represents a direction of the compass.  The compass is basically a for-loop going from ```minDegrees``` to ```maxDegrees. ```actualDegree``` is the index of this for loop. ```minDegrees``` and ```maxDegrees``` are the heading of the user ```-+ the field of view of the camera divided by 2``` which corresponds to what the user can see. 
The heading of the user is calculated in ```compass.java```. This value is always between ```0°``` and ```360°```. This means that ```actualDegree``` will always be between ```-x``` and ```360 + x``` where ```x``` is the field of view of the camera divided by 2.
Regarding your commit, if the ```actualDegree``` would be ```-10°``` for example and you take the absolute value this would be ```10°```. However, ```-10°``` and ```10°``` are not the same. If you now do ```-10 + 360``` you get ```350°``` and ```-10°``` is equal to ```350°```.

giommok (approved):

PR #167, opened 2021-04-27
Fix bug compass 180

magangloff:
Fixed bug #166 that was due to the low pass filter. When at around 180 degrees, the lowpass filter would calculate a weighted average. However the values were -pi and pi which would result in a wrong angle. This has been resolved creating a new class which expresses the angles in radians in their respective sine and cosine.

### DIFF
--- a/app/src/main/java/com/github/bgabriel998/softwaredevproject/utils/AngleLowpassFilter.java
+++ b/app/src/main/java/com/github/bgabriel998/softwaredevproject/utils/AngleLowpassFilter.java
@@ -1,0 +1,45 @@
+package com.github.bgabriel998.softwaredevproject.utils;
+
+import java.util.ArrayDeque;
+
+/**
+ * Lowpass filter for angles. Converts radians into sum of sine and cosine to overcome the problem
+ * that angle values are cyclic. For example when the angles changes from 2*pi to 0 or from -pi to pi
+ * See https://stackoverflow.com/a/18911252
+ */
+public class AngleLowpassFilter {
+    //Defines the length of the queue
+    private static final int LENGTH = 5;
+
+    private float sumSin, sumCos;
+
+    private final ArrayDeque<Float> queue = new ArrayDeque<>();
+
+    /**
+     * Adds an element to the queue and expresses the angle through their sine and cosine values
+     * @param radians angle that gets added
+     */
+    public void add(float radians){
+        sumSin += (float) Math.sin(radians);
+        sumCos += (float) Math.cos(radians);
+        queue.add(radians);
+
+        //If the queue is greater then the defined length
+        if(queue.size() > LENGTH){
+            //remove the oldest element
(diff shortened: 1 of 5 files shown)


app/src/test/java/com/github/bgabriel998/softwaredevproject/utils/AngleLowpassFilterTest.java
```diff
@@ -0,0 +1,57 @@
...
+        lowpassFilter.add(0);
+        assertEquals(0, lowpassFilter.average(), 0.01);
+
+        lowpassFilter.add((float) Math.PI/2);
+        assertEquals(Math.PI/4, lowpassFilter.average(), 0.01);
+    }
+
+    //Tests that the oldest element of the queue gets removed
+    @Test
+    public void lowpassFilterFullTest() {
```
giommok, line 23: If I understood correctly, in this test you add 6 values because the constant `LENGTH` in the filter is set to 5, so you can test that after you added `LENGHT + 1` angles the filter has removed one angle.
But what if we decide to increase the constant? I feel like we should modify this test using a for and the constant `LENGTH`, so it remains consistent even if we change `LENGTH`. Tell me what you think!

giommok (changes requested):
Well done and very nice comments, it was really easy to understand your code!
I'm not sure about a test, let me know what you think about the comment.

app/src/main/java/com/github/bgabriel998/softwaredevproject/utils/AngleLowpassFilter.java
```diff
@@ -0,0 +1,45 @@
+package com.github.bgabriel998.softwaredevproject.utils;
+
+import java.util.ArrayDeque;
+
+/**
+ * Lowpass filter for angles. Converts radians into sum of sine and cosine to overcome the problem
+ * that angle values are cyclic. For example when the angles changes from 2*pi to 0 or from -pi to pi
+ * See https://stackoverflow.com/a/18911252
```
ravifrancesco, line 8:
```suggestion
 * See <a href="StackOverflow">https://stackoverflow.com/a/18911252</a>
```

app/src/test/java/com/github/bgabriel998/softwaredevproject/utils/AngleLowpassFilterTest.java
ravifrancesco: Agreed, is it possible to make this a for/while loop to make it more maintainable?

magangloff: Yes, I agree, I just updated the test with values that are also better to understand. The test now increments the value every time by PI/4 so that the average can be simply calculated by hand.

ravifrancesco (approved):

giommok (approved):

PR #169, opened 2021-04-28
Image zoom

### LINKED ISSUE #143: #92 Full screen image zoom by pinch
Implement so that you can zoom an image by pinching
https://www.androidhive.info/2013/09/android-fullscreen-image-slider-with-swipe-and-pinch-zoom-gestures/

aolsson711:
Created Pinch zoom on fullscreen images from gallery. 
However impossible to test... 
closes #143


giommok (changes requested):
Code looks really good and really nice documentation, well done.
Anyway, I think we could add some tests. Further discussion on Discord.

app/src/main/java/com/github/bgabriel998/softwaredevproject/gallery/ZoomableImageView.java
```diff
@@ -0,0 +1,311 @@
...
+            maxTrans = 0;
+        }
+
+        if (trans < minTrans) {
+            return minTrans - trans;
+        }
+        if (trans > maxTrans) {
+            return maxTrans - trans;
+        }
+        return 0;
```
ravifrancesco, line 192:
```suggestion
        return trans < minTrans ? minTrans - trans : maxTrans - trans;
```

ravifrancesco (changes requested):
Small comment but overall the code looks good. Is there a way to keep track of the image size? In case could we test the zoom feature simulating a gesture and seeing if the scale increases/decreases

app/src/main/java/com/github/bgabriel998/softwaredevproject/gallery/ZoomableImageView.java
aolsson711: Not sure this is correct. I am pretty sure that there exists cases where (for easy numbers) minTrans = 0, maxTrans = 10, trans = 7. Then we want to return 0. not 3

giommok (changes requested):
Well done with the tests! 
I think we should use the upper-case convention for the constants in `ImageActivityTest`.
Would it be a problem to refactor them from Android Studio?

app/src/main/java/com/github/bgabriel998/softwaredevproject/gallery/ZoomableImageView.java
ravifrancesco: Oh yep you're right

ravifrancesco (approved):
Great job with the tests. I'll approve it, but consider @giommok request for the constant

giommok (approved):
All good!

PR #170, opened 2021-04-29
Filter overlapping mountains

magangloff:
Added a filter on the labeledPOIPoints. If two POIPoints are within +- 3 degrees, then only the highest POIPoint will be displayed. If a POIPoint is in the line of sight, then it is prioritised to the one that is not in the line of sight independent of the altitude.

TODO: Add settings to let the user choose how he wants to have the POIPoints displayed.


app/src/main/java/ch/epfl/sdp/peakar/points/ComputePOIPoints.java
```diff
@@ -0,0 +1,135 @@
...
+
+    /**
+     * Filters the labeled POIPoints and takes only the highest s.t. there are no other markers
+     * within 6° horizontally. POIPoints in the line of sight are prioritised. If a POIPoint is in
+     * the line of sight but has a smaller altitude then another POIPoint that is close but that is
+     * not in the line of sight, then only the POIPoint that is in the line of sight will be added
+     * @param labeledPOIPoints labeled POIPoints that get sorted
+     * @return Map<POIPoint, Boolean> of the highest filtered POIs
+     */
+     static Map<POIPoint, Boolean> filterHighestPOIs(Map<POIPoint, Boolean> labeledPOIPoints) {
```
aolsson711, line 91: Why is it only important to check the horizontal degrees and not vertical. 
Won't it become issues if you are near mountains. and see two mountains one smaller near you and one larger behind the smaller one. So you actually see both mountains. Will only the label for the one behind show?

app/src/main/res/layout-land/activity_image.xml
```diff
@@ -0,0 +1,22 @@
...
+        android:layout_height="match_parent"
+        app:layout_constraintTop_toTopOf="parent"
+        app:layout_constraintBottom_toBottomOf="parent"
+        app:layout_constraintEnd_toEndOf="parent"
+        app:layout_constraintStart_toStartOf="parent"
+        android:scaleType="centerCrop"
+        android:adjustViewBounds="true"
+        android:contentDescription="@string/description_gallery_fullscreen_image"/>
+
+</androidx.constraintlayout.widget.ConstraintLayout>
```
aolsson711, line 22: This is not used anymore. Remove

aolsson711 (commented):
Code overall looks good. Nice Work! 
Question regarding why only horizontal is important not vertical. 
And to remove an extra layout that is not necessary

app/src/main/java/ch/epfl/sdp/peakar/points/ComputePOIPoints.java
ravifrancesco: I do agree, although I think this needs some real-world tests as it is difficult to predict the various situations that might occur.

ravifrancesco (approved):

aolsson711 (approved):
Great Work!

PR #188, opened 2021-05-01
Camera methods to utility

magangloff:
Addressed some issues pointed out by the TAs:
- Created CameraUtility class that contains methods for the camera
- Added Javadoc for the classes in the camera package
- Renamed IF to Interface and extracted topography interface

### DIFF
--- a/app/src/main/java/ch/epfl/sdp/peakar/utils/CameraUtilities.java
+++ b/app/src/main/java/ch/epfl/sdp/peakar/utils/CameraUtilities.java
@@ -10,6 +10,9 @@ import android.util.SizeF;
 import androidx.camera.core.AspectRatio;
 import androidx.core.util.Pair;
 
+/**
+ * Utility class for the Camera Activity, contains utility methods for the camera-preview and compass
+ */
 public final class CameraUtilities {
 
     /**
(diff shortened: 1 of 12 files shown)


aolsson711 (approved):
Code is just moving functions, No comments! It all looks good

app/src/main/java/ch/epfl/sdp/peakar/utils/CameraUtilities.java
```diff
@@ -0,0 +1,111 @@
...
+ */
+public final class CameraUtilities {
+
+    /**
+     * Calculate the aspect ratio of the display in function of the width and height of the screen
+     * @param width width of the preview in Pixels
+     * @param height height of the preview in Pixels
+     * @return Aspect ratio of the phone
+     */
+    public static int aspectRatio(int width, int height){
```
ravifrancesco, line 24: What happens if the actual aspect ratio of the phone is not near this ratio? For example in the case of 21:9?

magangloff: Then it will return the nearest aspectRatio which is 16:9. CameraX only supports 4:3 and 16:9 for now but the camera will still be displayed in fullscreen mode but the picture that is taken without the UI will be in 16:9 aspect ratio.

ravifrancesco: Okay nice thanks!

ravifrancesco (approved):

PR #189, opened 2021-05-01
Add POI caching

### LINKED ISSUE #185: Saving POIs in the cache

### LINKED ISSUE #186: Retrieve POIs from the cache

bgabriel998:
Add poi caching to `ComputePOIPoint.java` file.
The class POICache handles the serialization and deserialization of the List of POI and bounding box. 
The serialized object is from type `POICacheContent`. 
The file is saved under cache folder of the android device. 

**Important** : The download is triggered when the user location is not included in the middle of the bounding box. This means in a bounding box from size 5km² inside the cached bounding box. 

Close #185 
Close #186


app/src/androidTest/java/ch/epfl/sdp/peakar/points/POICacheTest.java
```diff
@@ -0,0 +1,147 @@
...
+        //Create new cache file
+        poiCache.savePOIDataToCache(inputArrayList,boundingBox,context);
+        ArrayList<POIPoint> result = poiCache.getCachedPOIPoints(context);
+
+        assertEquals(inputArrayList.size(), result.size());
+        assertTrue(inputArrayList.contains(result.get(0)));
+        assertTrue(inputArrayList.contains(result.get(1)));
+        assertTrue(inputArrayList.contains(result.get(2)));
+        assertTrue(inputArrayList.contains(result.get(3)));
+    }
```
aolsson711, line 132: I am a bit petty. But when I think about this test. Say that both the delete function and the add function doesn't work. So you never remove the item. (and also never create it) then the test still passes right? 
That should be fixed. Maybe check array list before adding the item again. Or splitting it to several tests.

app/src/main/java/ch/epfl/sdp/peakar/general/InitActivity.java
```diff
@@ -54,8 +60,8 @@ protected void onCreate(Bundle savedInstanceState) {
      * Init application global stuff before opening the main menu
      */
     private synchronized void initApp(){
-        FirebaseAccount firebaseAccount = FirebaseAccount.getAccount();
-        firebaseAccount.synchronizeUserProfile();
+        //FirebaseAccount firebaseAccount = FirebaseAccount.getAccount();
+        //firebaseAccount.synchronizeUserProfile();
```
aolsson711, line 64:
```suggestion
```

app/src/main/java/ch/epfl/sdp/peakar/points/ComputePOIPoints.java
```diff
@@ -37,17 +40,51 @@
...
+     * or from provider. If the cached data corresponds to
+     * the user location, no download is made and the POIs
+     * are retrieved from cached file
+     * @param userPoint user location
      */
     private static void getPOIs(UserPoint userPoint){
+        //Retrieve cache instance
+        POICache poiCache = POICache.getInstance();
+        //Check if the user is in the cached BB. If yes
+        //load data from cache / if not download data
```
aolsson711, line 62:
```suggestion
```

Thinking about the review we got from TA to remove these type of comments. I think it is well explained in the javadoc

app/src/main/java/ch/epfl/sdp/peakar/points/ComputePOIPoints.java
```diff
@@ -37,17 +40,51 @@
...
+     * @param userPoint location of the user
+     */
+    private static void getPOIsFromCache(UserPoint userPoint){
+        ArrayList<POIPoint> cachedPOIs = POICache.getInstance().getCachedPOIPoints(ctx);
+        POIPoints.addAll(cachedPOIs.stream().peek(poiPoint ->
+        {
+            poiPoint.setHorizontalBearing(userPoint);
+            poiPoint.setVerticalBearing(userPoint);
+        }).collect(Collectors.toList()));
+        //getLabeledPOIs(userPoint);
```
aolsson711, line 80:
```suggestion
```

if it is something for the future add TODO instead

app/src/main/java/ch/epfl/sdp/peakar/points/POICache.java
```diff
@@ -0,0 +1,155 @@
...
+        String serializesCache = gson.toJson(poiCacheContent);
+        saveJson(serializesCache,context);
+    }
+
+    /**
+     * Save the serialized JSON to text file
+     * @param serializedObject string in json format
+     * @param context context
+     */
+    private static void saveJson(String serializedObject,Context context){
```
aolsson711, line 68: Are we going to have more cached objects that will use Json? In that case we should move this to a class for itself.

app/src/main/java/ch/epfl/sdp/peakar/points/POICache.java
```diff
@@ -0,0 +1,155 @@
...
+            Log.e("Exception", "File write failed cache POIPoints: " + e.toString());
+        }
+    }
+
+
+    /**
+     * Read text file and retrieve serialized
+     * @return serialized string
+     */
+    private static String readJSON(Context context){
```
aolsson711, line 87: If comment regarding saveJson applies this function should also be moved.

app/src/main/java/ch/epfl/sdp/peakar/points/POICache.java
```diff
@@ -0,0 +1,155 @@
...
+        return fileContent.toString();
+    }
+
+    /**
+     * Retrieve POI data from cache and overwrite cachedPOIPoints and cachedBoundingBox
+     * with cache file values
+     */
+    private static void retrievePOIDataFromCache(Context context){
+        String fileContent = readJSON(context);
+        Gson gson = new Gson();
```
aolsson711, line 110: I am not an expert at Gson. But I see that in several function we do "new Gson()" is that something that is possible to do in the constructor once and then use the same Gson object?

app/src/main/java/ch/epfl/sdp/peakar/points/POICache.java
```diff
@@ -0,0 +1,155 @@
...
+        return cachedBoundingBox.contains(userPoint.getLatitude(), userPoint.getLongitude());
+    }
+
+
+    /**
+     * Checks if cache file is present in device Cache folder
+     * @param context application context
+     * @return true if file is present. False if no file found
+     */
+    public boolean isCacheFilePresent(Context context){
```
aolsson711, line 140: This function should in that case also be moved.

aolsson711 (commented):
Code looks nice. Good work! 
Some comments regarding a test and if the cache part should be a function on its own.

app/src/main/java/ch/epfl/sdp/peakar/general/InitActivity.java
bgabriel998: This was for debugging purpose. I will add them again in next commit

app/src/main/java/ch/epfl/sdp/peakar/points/ComputePOIPoints.java
bgabriel998: You are right

app/src/main/java/ch/epfl/sdp/peakar/points/ComputePOIPoints.java
bgabriel998: Yes indeed that is something for a later task, I will add a todo instead of removing the comment completely

app/src/main/java/ch/epfl/sdp/peakar/points/POICache.java
bgabriel998: I think the caching of the 3D map will be done in the same class so this method can stay here

app/src/main/java/ch/epfl/sdp/peakar/points/POICache.java
bgabriel998: Same as saveJSON -> no moving

app/src/main/java/ch/epfl/sdp/peakar/points/POICache.java
bgabriel998: Gson is a library I used to serialize Objects directly in JSON format, I followed the lib tuto for this method

app/src/main/java/ch/epfl/sdp/peakar/points/POICache.java
bgabriel998: No, this is really usefull to check if the data has to be download or retrieved from the cache.

app/src/androidTest/java/ch/epfl/sdp/peakar/points/POICacheTest.java
bgabriel998: I think I will rework the tests... some are still problematic

app/src/main/java/ch/epfl/sdp/peakar/points/POICache.java
aolsson711: Yes I meant if all cache should be moved but since that is not the case the comment no longer applies.

app/src/main/java/ch/epfl/sdp/peakar/points/POICache.java
aolsson711: Hmm.. Just I am not sure a new one needs to be created everytime. But as I said I am not sure...

giommok: https://sites.google.com/site/gson/gson-user-guide

> The Gson instance does not maintain any state while invoking Json operations. So, you are free to reuse the same object for multiple Json serialization and deserialization operations.

It seems like we can indeed use only one instance!

app/src/main/java/ch/epfl/sdp/peakar/points/POICache.java
giommok, line 68: I think there is a much easier way to save data to a file from a JSON formatted string with GSON:
`Gson gson = new Gson();
gson.toJson(yourContent, new FileWriter(yourfilePath));`
Check [here](https://www.baeldung.com/gson-save-file) for more detailed information.

app/src/main/java/ch/epfl/sdp/peakar/points/POICache.java
```diff
@@ -0,0 +1,155 @@
...
+     * @param userPoint user location
+     * @return true if the user is in the middle of the cached bounding box
+     *          false if not
+     */
+    public boolean isUserInBoundingBox(UserPoint userPoint,Context context){
+        //Get the bounding box from file if not already present
+        if(cachedBoundingBox == null)
+            retrievePOIDataFromCache(context);
+        //Decrease the size of the bounding box by half
+        //BoundingBox innerBox = cachedBoundingBox.increaseByScale(INNER_BOUNDING_BOX_SCALING_FACTOR);
```
giommok, line 130: This line is commented. Is it an error or was it on purpose? If so, can you add a comment explaining why (for example, if we can use it in future)?

app/src/main/java/ch/epfl/sdp/peakar/points/POICache.java
giommok, line 87: I'm not sure it is convenient in this case, but GSON offers a "direct" way to read from files.
Check [here](https://stackoverflow.com/questions/29965764/how-to-parse-json-file-with-gson) for more information.

app/src/main/java/ch/epfl/sdp/peakar/points/POICacheContent.java
```diff
@@ -0,0 +1,40 @@
+package ch.epfl.sdp.peakar.points;
+
+import org.osmdroid.util.BoundingBox;
+
+import java.util.ArrayList;
+
+/**
+ * Class used for serialization and deserialization
```
giommok, line 8: Maybe it's just me, but this comment made me think that this class contains the method we can use for JSON serialization and deserialization, whereas it just represents the content that is serialized and deserialized to JSON using other methods.
What do you think?

giommok (changes requested):
Nice job!
I left some comments about GSON usage and a few minor things.

app/src/main/java/ch/epfl/sdp/peakar/general/InitActivity.java
```diff
@@ -43,6 +44,12 @@ protected void onCreate(Bundle savedInstanceState) {
                 //Add here permissions
         });
 
+        try {
+            Thread.sleep(3000);
```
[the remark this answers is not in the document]
bgabriel998: The Sleep has been added so the Activity has time to display all the permissions that the user needs to accept. I will open an issue to find a better solution. Without the sleep, the permissions are asked when the user leaves the app...

app/src/main/java/ch/epfl/sdp/peakar/points/POICache.java
bgabriel998: Simplified in next commit. Good suggestion !

app/src/main/java/ch/epfl/sdp/peakar/points/POICache.java
bgabriel998: Uncommented in next commit

app/src/main/java/ch/epfl/sdp/peakar/points/POICache.java
bgabriel998: Corrected in next commit

app/src/main/java/ch/epfl/sdp/peakar/points/POICacheContent.java
bgabriel998: Right, i'll correct that

app/src/main/java/ch/epfl/sdp/peakar/points/POICache.java
bgabriel998: Update --> The ReadJSON method is back since i have issue to read/write using GSON methods

app/src/main/java/ch/epfl/sdp/peakar/points/POICache.java
bgabriel998: Ok i'll change that

giommok (approved):

app/src/androidTest/java/ch/epfl/sdp/peakar/TestingConstants.java
```diff
@@ -50,4 +50,5 @@
 
     //Cache constants
     public static final String CACHE_FILE_NAME_TEST = "POIPointsCache.txt";
+    public static final String CACHE_FILE_PATH_TEST = "";
```
aolsson711, line 53: When is this used? Do we need an empty constant? Or can this be refactored to not be used at all?

aolsson711 (commented):
Everything looks good. Super small comment/question regarding an empty constant

app/src/androidTest/java/ch/epfl/sdp/peakar/TestingConstants.java
bgabriel998: You are right, that is some rest from my intense debug session

aolsson711 (approved):
Good Work!

PR #214, opened 2021-05-09
Update camera UI

magangloff:
Updated UI of the camera activity to the new UI:
- Added profile button. If the user is not signed in, then the google sign authentification will be launched. Else the ProfileActivity will be launched.
- Changed image of button to take pictures and button to change display modes.
- By default, the old compass will be removed and a new compass will be displayed beside the take picture button. This new compass is used to toggle between displaying or not the old compass. A setting has also been added for this.


app/src/androidTest/java/ch/epfl/sdp/peakar/camera/CameraActivityTest.java
```diff
@@ -36,11 +50,39 @@
...
         new ComputePOIPoints(context);
+        sharedPreferences = PreferenceManager.getDefaultSharedPreferences(context);
+        SharedPreferences.Editor editor = sharedPreferences.edit();
+        editor.clear().commit();
+
+        /* Make sure no user is signed in before tests */
+        AuthService.getInstance().signOut(InstrumentationRegistry.getInstrumentation().getTargetContext());
+        /* Create a new one */
+        registerAuthUser();
+        String user1 = (USERNAME_CAMERA + AuthService.getInstance().getID()).substring(0, Account.NAME_MAX_LENGTH - 1);
```
giommok, line 69:
```suggestion
```
I think the string is never used, maybe we can remove it at all.

app/src/androidTest/java/ch/epfl/sdp/peakar/camera/CameraActivityTest.java
```diff
@@ -55,27 +97,50 @@ public void cleanUp(){
...
+    }
+
+    public void createTestUser() {
+        if(FirebaseAuth.getInstance().getCurrentUser() == null) {
+            AuthService.getInstance().signOut(InstrumentationRegistry.getInstrumentation().getTargetContext());
+            registerAuthUser();
+        }
+        else {
+            FirebaseAuthService.getInstance().forceRetrieveData();
         }
```
giommok, line 143: In my PR I've put these lines inside the `registerAuthUser()` method as they were repeated everywhere. As we don't know which PR will be merged first, should we add here a TODO comment saying that we can replace all these lines with just a `registerAuthUser()`?

app/src/androidTest/java/ch/epfl/sdp/peakar/user/ProfileActivityTest.java
```diff
@@ -47,7 +47,7 @@
 import static org.junit.Assert.assertSame;
 import static org.junit.Assert.fail;
 
-public class ProfileActivityTest {
+public class    ProfileActivityTest {
```
giommok, line 50:
```suggestion
public class ProfileActivityTest {
```
I think you accidentally added some blank spaces here.

app/src/main/java/ch/epfl/sdp/peakar/camera/CameraActivity.java
```diff
@@ -173,6 +185,8 @@ public void takePictureListener(View view) throws IOException {
         Bitmap bitmap = CameraUtilities.combineBitmaps(cameraBitmap, compassBitmap);
         //Store the bitmap on the user device
         StorageHandler.storeBitmap(this, bitmap);
+        //Set visibility to invisible again after 100ms
```
giommok, line 188:
```suggestion
        //Set visibility to invisible again after FLASH_TIME_MS
```
I think this comment does not match the actual sleep time and could be confusing in future if we expect a 100ms time delay but it actually waits just 5ms. Maybe we should just leave the constant so one can go up and see the actual value.

giommok (changes requested):
Looks good to me! 
Check the comments for some small changes.

app/src/androidTest/java/ch/epfl/sdp/peakar/camera/CameraActivityTest.java
magangloff: Yes, I added TODO comments.

giommok (approved):

PR #229, opened 2021-05-13
Loading bar

ravifrancesco:
The `SettingsMapActivity` has now a new GUI, using the toolbar and the color palette used in the app.

When a long-press is performed, the touch will be disabled during the download, and a loading circle will be displayed. When the download is terminated the user will be prompted to the `SettingActiivty.`


app/src/androidTest/java/ch/epfl/sdp/peakar/general/SettingsMapActivityTest.java
```diff
@@ -59,19 +58,26 @@ public void cleanUp(){
     /* Test that the "OK" button appears after adding a pin to the map */
```
giommok, line 58: I think this comment is outdated looking at the changes, is it correct?

app/src/androidTest/java/ch/epfl/sdp/peakar/general/SettingsMapActivityTest.java
```diff
@@ -115,12 +121,12 @@ public void testBackButton() {
 
     /* Tests that the activity is terminated after the ok button is pressed. */
```
giommok, line 122: Comment outdated.

giommok (changes requested):
Seems good to me.
However, I think there are some tests descriptions that need to be updated, check my comments, please.

giommok (approved):

magangloff (commented):
looks all good and much cleaner than before. Just a question, what happens if the user uses his back button to exit the download, will the download be cancelled or continued in the background?

ravifrancesco:
> looks all good and much cleaner than before. Just a question, what happens if the user uses his back button to exit the download, will the download be cancelled or continued in the background?

You cannott press the back button, as touchscreen is disabled during the download

magangloff:
> > looks all good and much cleaner than before. Just a question, what happens if the user uses his back button to exit the download, will the download be cancelled or continued in the background?
> 
> You cannot press the back button, as touchscreen is disabled during the download

Well yes, you actually can, by using your back buttons from the phone(like swipe gestures for example or the android back buttons)

ravifrancesco:
> > > looks all good and much cleaner than before. Just a question, what happens if the user uses his back button to exit the download, will the download be cancelled or continued in the background?
> > 
> > 
> > You cannot press the back button, as touchscreen is disabled during the download
> 
> Well yes, you actually can, by using your back buttons from the phone(like swipe gestures for example or the android back buttons)

Oh I see, I will work on that as a task for the next sprint there

magangloff (approved):
Now, the user is forced to stay on the download activity when he starts a download. I'm not sure if this should be handled like that.

ravifrancesco:
> Now, the user is forced to stay on the download activity when he starts a download. I'm not sure if this should be handled like that.

I think it should be like this: this way the user knows when the download is running (so the user will not close the app or try to change settings while the download is running)

PR #244, opened 2021-05-16
Profile with data

giommok:
- Now the new profile activity shows the correct image profile, collected mountains and user info (score, username).
- In addition, you can sign out and you can change the username from the new activity too.
- Modified the profile launcher created by @magangloff to start the new profile activity (so you can try the new profile UI by going to the camera and pressing the profile button).

TODO: add the country flags, set the correct points of the peaks and add sorting options.
Also, we still have to implement the profile features of a non authenticated user (like add friend).


app/src/main/java/ch/epfl/sdp/peakar/general/InitActivity.java
```diff
@@ -12,6 +12,7 @@
 import androidx.core.content.ContextCompat;
 
 import com.google.firebase.FirebaseApp;
+import com.google.firebase.database.FirebaseDatabase;
```
magangloff, line 15: Unused import can be removed.
```suggestion

```

app/src/main/java/ch/epfl/sdp/peakar/user/profile/NewProfileActivity.java
```diff
@@ -2,23 +2,32 @@
 
 import androidx.appcompat.app.AppCompatActivity;
 
+import android.content.Intent;
 import android.graphics.PorterDuff;
 import android.graphics.PorterDuffColorFilter;
 import android.graphics.drawable.Drawable;
+import android.net.Uri;
 import android.os.Bundle;
+import android.text.InputType;
```
magangloff, line 11: Unused import can be removed.
```suggestion

```

app/src/main/java/ch/epfl/sdp/peakar/user/profile/NewProfileActivity.java
```diff
@@ -2,23 +2,32 @@
 
 import androidx.appcompat.app.AppCompatActivity;
 
+import android.content.Intent;
 import android.graphics.PorterDuff;
 import android.graphics.PorterDuffColorFilter;
 import android.graphics.drawable.Drawable;
```
magangloff, line 8: Unused import can be removed.
```suggestion

```

app/src/main/java/ch/epfl/sdp/peakar/user/profile/NewProfileActivity.java
```diff
@@ -28,20 +37,51 @@
...
+        isAuthProfile = startingIntent.getBooleanExtra(AUTH_INTENT, false);
+
+        if(!isAuthProfile) {
+            String otherId = startingIntent.getStringExtra(OTHER_INTENT);
+            new Thread(() -> {
+                otherAccount = OtherAccount.getInstance(otherId);
+                runOnUiThread(() -> {
+                    setContentView(R.layout.activity_new_profile);
+
+                    StatusBarHandler.StatusBarTransparent(this);
```
aolsson711, line 65: These two are in both if and else. Is it needed because of the  `runOnUiThread()` or could they be preformed before the if condition?

app/src/main/java/ch/epfl/sdp/peakar/user/profile/NewProfileActivity.java
```diff
@@ -80,27 +122,30 @@ private void hideUI(boolean self, boolean friends) {
...
-                        false, 0, 0, "2000-01-01"),
-                new NewCollectedItem("Kebnekaise", 710000, 2097,
-                        true, 0, 0, "2000-01-01")
-        ));
-        //ArrayList<NewCollectedItem> items = new ArrayList<>();
+        ArrayList<NewCollectedItem> newItems = new ArrayList<>();
+        for(POIPoint discoveredPeak: AuthService.getInstance().getAuthAccount().getDiscoveredPeaks()) {
+            NewCollectedItem newCollectedItem = new NewCollectedItem(
+                    discoveredPeak.getName(),
+                    0,
```
aolsson711, line 132: is the points not saved currently in `discoveredPeak`?

app/src/main/java/ch/epfl/sdp/peakar/user/profile/NewProfileActivity.java
```diff
@@ -80,27 +122,30 @@ private void hideUI(boolean self, boolean friends) {
...
-        ArrayList<NewCollectedItem> items = new ArrayList<>(Arrays.asList(
-                new NewCollectedItem("Mont Blanc - Monte Bianco", 4810000, 4810,
-                        true, 0, 0, "2000-01-01"),
-                new NewCollectedItem("Diablerets", 1210000, 3210,
-                        false, 0, 0, "2000-01-01"),
-                new NewCollectedItem("Kebnekaise", 710000, 2097,
-                        true, 0, 0, "2000-01-01")
-        ));
-        //ArrayList<NewCollectedItem> items = new ArrayList<>();
+        ArrayList<NewCollectedItem> newItems = new ArrayList<>();
```
aolsson711, line 128:
```suggestion
        ArrayList<NewCollectedItem> items = new ArrayList<>();
```

app/src/main/java/ch/epfl/sdp/peakar/user/profile/NewProfileActivity.java
```diff
@@ -131,4 +176,78 @@ private void expandSelected(boolean expand) {
...
+     * On change username button click
+     */
+    public void changeUsernameButton(View view) {
+        // Hide username
+        findViewById(R.id.profile_username).setVisibility(View.GONE);
+
+        // Show edit text
+        findViewById(R.id.profile_username_edit).setVisibility(View.VISIBLE);
+
+        // Change listener so next click will result in a username chance
```
aolsson711, line 198:
```suggestion
        // Change listener so next click will result in a username change
```

app/src/main/java/ch/epfl/sdp/peakar/user/profile/NewProfileActivity.java
```diff
@@ -131,4 +176,78 @@ private void expandSelected(boolean expand) {
...
+        findViewById(R.id.profile_change_username).setOnClickListener(this::confirmUsernameButton);
+    }
+
+    /**
+     * On confirm username change button click
+     */
+    public void confirmUsernameButton(View view) {
```
aolsson711, line 205: I tested the change username functionality. 
Is it possible to change a bit of how it works. 

Currently when you press the change username button

1. It changes to an edit text view that looks different from the normal text view (other colors smaller text etc.).
2. You press the new text view to type in your name
3. You press the change username button again to do the change.

Would it be possible to change this flow to, when you press the button:

1. The view changes to edit text (that looks just like the username text, and you get the keyboard directly, to type in a new username. 
2. When you press enter on the keyboard it changes name, if you press anywhere else on the screen it stops trying to change username? 

I think that flow would make more sense and it removes 2 extra clicks for the user

aolsson711 (commented):
Good code. some remark on naming etc. 

Biggest question is regarding the flow of changing username, see that comment and we can talk more on Discord about it.

app/src/main/res/layout/activity_new_profile.xml
```diff
@@ -12,14 +12,15 @@
         android:id="@+id/profile_sign_out"
```
magangloff, line 12: On the NewProfileActivity the friend's button is invisible if the user is looking at his profile. The only way to see his friends will be through the rankings if I'm not mistaking, should we also add the possibility to display a list of friends accessible from the user profile?

app/src/androidTest/java/ch/epfl/sdp/peakar/user/NewProfileActivityTest.java
```diff
@@ -0,0 +1,223 @@
+package ch.epfl.sdp.peakar.user;
+
+import android.app.Activity;
```
magangloff, line 3: Unused imports can be removed.

app/src/main/java/ch/epfl/sdp/peakar/user/profile/NewProfileActivity.java
```diff
@@ -80,27 +122,30 @@ private void hideUI(boolean self, boolean friends) {
 
     /**
      * Fill list view.
-     * TODO get list from DB?
+     * TODO show correct points and correct date
```
magangloff, line 125: I think that the correct country should also be filled in this method, maybe add it in the TODOs.

magangloff (changes requested):
The new profile looks really nice!
Suggested a few changes.

app/src/main/java/ch/epfl/sdp/peakar/user/profile/NewProfileActivity.java
magangloff: The color is actually the same for me, but I agree that it would be nice for the keyboard to appear when clicking the first time on the button and to have an option to cancel the username change (like pressing anywhere else on the screen). I think that this is a "nice to have" so maybe you shouldn't sacrifice too much time on this right now if you can't find a fast way to implement these features.

magangloff (changes requested):
Added suggestions for changes made by Alex:
- Keyboard opens on the first button press
- Possibility to change the username when clicking on entering on the keyboard
- Make keyboard disappear when clicking somewhere else on the scree when entering the username (TODO: revert to old username after cancelling the username change)

app/src/main/java/ch/epfl/sdp/peakar/user/profile/NewProfileActivity.java
magangloff: Suggested some changes below to handle these points pointed out by Alex.

magangloff (changes requested):
When a user selects a new Gmail, then the profile activity will be displayed with the username @ and the collection will display Empty collection (see picture below). Maybe if the username was not set yet by the user, the keyboard should directly be opened s.t. the user can change his username directly. if he cancels, then maybe display a text like: "You need to change the username to save the discovered peaks" instead of "Empty Collection".

Screenshot_20210517-122422

app/src/main/res/layout/activity_new_profile.xml
giommok: I think we decided to not have it there as we want all the social part to be in the social activity.

app/src/main/java/ch/epfl/sdp/peakar/user/profile/NewProfileActivity.java
giommok: Good point

app/src/main/java/ch/epfl/sdp/peakar/user/profile/NewProfileActivity.java
giommok: Refactored from AndroidStudio.

app/src/main/java/ch/epfl/sdp/peakar/user/profile/NewProfileActivity.java
giommok: Exactly.

app/src/main/java/ch/epfl/sdp/peakar/user/profile/NewProfileActivity.java
giommok: Yep. As in the case of another user we need to load its data first, I prefer not to show the layout before the data has been loaded.
What do you think?

app/src/main/java/ch/epfl/sdp/peakar/user/profile/NewProfileActivity.java
giommok: Thanks to you both for the suggestion (and for the solution @magangloff !).
Now, when you press on the username change button, the edit text and the keyboard are shown, whereas the button and the username text view are set as gone. 
When you press enter on the keyboard the change is performed (or denied, if necessary). 

Also, with the solution proposed by @magangloff we solved also the problem of how to force a username change for new users (in other words, how to handle the registration phase). 
Basically, we force new users to insert something in the edit text and even if you try to press somewhere else on the screen, nothing changes.

You can test again the functionality and tell me if now it is good for you.

aolsson711 (approved):
Nice work! 
The change username part also became really good, well done!

app/src/main/java/ch/epfl/sdp/peakar/user/profile/NewProfileActivity.java
```diff
@@ -28,20 +40,59 @@
...
+
+            // If the user is not registered, force a username change
+            if(AuthService.getInstance().getAuthAccount().getUsername().equals(Account.USERNAME_BEFORE_REGISTRATION)) {
+                changeUsernameButton(null);
+                // Set text view
+                ((TextView)findViewById(R.id.profile_empty_text)).setText(R.string.empty_collection_not_registered);
+            } else {
+                fillListView();
+            }
+        }
```
magangloff, line 95: Add a listener for keyboard changes to handle the case when the keyboard is open and the user presses the back button.
```suggestion
        findViewById(R.id.profile_collection).setOnApplyWindowInsetsListener((v, insets) -> {
            if(!isKeyboardOpened()){
                setUiUsernameChanged();
            }
            return insets;
        });
```

app/src/main/java/ch/epfl/sdp/peakar/user/profile/NewProfileActivity.java
```diff
@@ -131,4 +192,119 @@ private void expandSelected(boolean expand) {
...
+        InputMethodManager imm = (InputMethodManager) getSystemService(Context.INPUT_METHOD_SERVICE);
+        imm.showSoftInput(changeUsername, InputMethodManager.SHOW_IMPLICIT);
+
+        changeUsername.setOnKeyListener((v, keyCode, event) -> {
+            if ((event.getAction() == KeyEvent.ACTION_DOWN) &&
+                    (keyCode == KeyEvent.KEYCODE_ENTER)) {
+                confirmUsernameButton();
+                return true;
+            }
+            return false;
```
magangloff, line 228: Return true for key_up and down button press.
```suggestion
            }
            return true;
```

app/src/main/java/ch/epfl/sdp/peakar/user/profile/NewProfileActivity.java
```diff
@@ -131,4 +192,119 @@ private void expandSelected(boolean expand) {
...
+        });
+    }
+
+    /**
+     * On confirm username change
+     */
+    public void confirmUsernameButton() {
+        EditText usernameEdit = ((EditText)findViewById(R.id.profile_username_edit));
+        hideKeyboard();
```
magangloff, line 237: Remove hideKeyboard();
```suggestion

```

app/src/main/java/ch/epfl/sdp/peakar/user/profile/NewProfileActivity.java
```diff
@@ -131,4 +192,119 @@ private void expandSelected(boolean expand) {
...
+            public void run() {
+                // Change the username and wait for the task to end
+                ProfileOutcome result = AuthService.getInstance().getAuthAccount().changeUsername(newUsername);
+
+                // Update the view on the UI thread
+                runOnUiThread(() -> {
+                    ((EditText)findViewById(R.id.profile_username_edit)).getText().clear();
+
+                    // If username has changed, update the username text view
+                    if(result == ProfileOutcome.USERNAME_CHANGED || result == ProfileOutcome.USERNAME_REGISTERED) ((TextView)findViewById(R.id.profile_username)).setText(AuthService.getInstance().getAuthAccount().getUsername());
```
magangloff, line 253: hide keyboard only if the user changed successfully his name or registered his account.
```suggestion
                    // If username has changed, hide the keyboard and update the username text view
                    if(result == ProfileOutcome.USERNAME_CHANGED || result == ProfileOutcome.USERNAME_REGISTERED) {
                        hideKeyboard();
                        ((TextView)findViewById(R.id.profile_username)).setText(AuthService.getInstance().getAuthAccount().getUsername());
                    }
```

app/src/main/java/ch/epfl/sdp/peakar/user/profile/NewProfileActivity.java
```diff
@@ -131,4 +192,119 @@ private void expandSelected(boolean expand) {
...
+                    snackbar.show();
+
+                    // Hide username edit
+                    findViewById(R.id.profile_username_edit).setVisibility(View.GONE);
+
+                    // Show username
+                    findViewById(R.id.profile_username).setVisibility(View.VISIBLE);
+
+                    // Show change button
+                    findViewById(R.id.profile_change_username).setVisibility(View.VISIBLE);
```
magangloff, line 267: Remove Ui change (handled when the keyboard layout changes)
```suggestion

```

app/src/main/java/ch/epfl/sdp/peakar/user/profile/NewProfileActivity.java
```diff
@@ -131,4 +192,119 @@ private void expandSelected(boolean expand) {
...
+                });
+            }
+        };
+        changeThread.start();
+    }
+
```
magangloff, line 273: Add method to update UI after that the user changed or cancelled his username change
```suggestion
    /**
     * Hides the keyboard and updates the UI after the username was changed or cancelled
     */
    private void setUiUsernameChanged() {
        // Hide username edit
        findViewById(R.id.profile_username_edit).setVisibility(View.GONE);

        // Show username
        findViewById(R.id.profile_username).setVisibility(View.VISIBLE);

        // Show change button
        findViewById(R.id.profile_change_username).setVisibility(View.VISIBLE);
    }
```

app/src/main/java/ch/epfl/sdp/peakar/user/profile/NewProfileActivity.java
```diff
@@ -131,4 +192,119 @@ private void expandSelected(boolean expand) {
...
+    /**
+     * Method to hide the keyboard
+     */
+    private void hideKeyboard() {
+        InputMethodManager imm = (InputMethodManager) getSystemService(Context.INPUT_METHOD_SERVICE);
+        imm.hideSoftInputFromWindow(getCurrentFocus().getWindowToken(), 0);
+    }
+
```
magangloff, line 281: Add method to detect if keyboard is opened or not.
```suggestion
    /**
     * Checks if the keyboard is opened or not by calculating the difference between the actual
     * height of the profile_collection ListView and the visible height of the View
     *
     * @return true if the keyboard is opened
     */
    private boolean isKeyboardOpened(){
        Rect visibleBounds = new Rect();
        ListView collectionView = findViewById(R.id.profile_collection);
        collectionView.getWindowVisibleDisplayFrame(visibleBounds);
        double heightDiff = collectionView.getHeight() - visibleBounds.height();
        return heightDiff > 0;
    }
```

app/src/main/java/ch/epfl/sdp/peakar/user/profile/NewProfileActivity.java
```diff
@@ -131,4 +192,119 @@ private void expandSelected(boolean expand) {
...
+            hideKeyboard();
+
+            // Hide username edit
+            findViewById(R.id.profile_username_edit).setVisibility(View.GONE);
+
+            // Show username
+            findViewById(R.id.profile_username).setVisibility(View.VISIBLE);
+
+            // Show change button
+            findViewById(R.id.profile_change_username).setVisibility(View.VISIBLE);
```
magangloff, line 294: Remove UI changes (handled by keyboard change listener)
```suggestion

```

magangloff (changes requested):
Suggested changes for a bug when the user was not registered and changes his username to an existing or not valid username. This would close the keyboard and display the old username which in this case would be the "@" since the user did not have any other username.
With the changes, the keyboard will stay open if the username is not valid. This would be the case every time, independent of the registration of the username but this can be changed if you prefer to only let the keyboard opened if the user is not registered and enters a wrong username

Also added a listener for when the back button of the device is pressed while the keyboard is opened. 

The suggestions will need some imports to be added.

magangloff (approved):

PR #248, opened 2021-05-18
Update of ranking tests

giommok:
#228 
I've found a nice way of checking that the `List` under a `ListView` is sorted, so I added a utility method in `SocialActivityTest` to do it and I implemented a test using this method.

As this task required only a small amount of time, I've also implemented the missing code to show the users' avatar in the `SocialActivity`.

### DIFF
--- a/app/src/androidTest/java/ch/epfl/sdp/peakar/social/SocialActivityTest.java
+++ b/app/src/androidTest/java/ch/epfl/sdp/peakar/social/SocialActivityTest.java
@@ -1,27 +1,96 @@
 package ch.epfl.sdp.peakar.social;
 
+import android.util.Log;
+import android.view.View;
+import android.widget.ListView;
+
 import androidx.test.espresso.intent.Intents;
 import androidx.test.espresso.matcher.ViewMatchers;
 import androidx.test.ext.junit.rules.ActivityScenarioRule;
 import androidx.test.ext.junit.runners.AndroidJUnit4;
 
+import org.hamcrest.Description;
+import org.hamcrest.Matcher;
+import org.hamcrest.TypeSafeMatcher;
 import org.junit.After;
+import org.junit.AfterClass;
 import org.junit.Before;
+import org.junit.BeforeClass;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.List;
+
 import ch.epfl.sdp.peakar.R;
+import ch.epfl.sdp.peakar.database.Database;
+import ch.epfl.sdp.peakar.database.DatabaseReference;
+import ch.epfl.sdp.peakar.user.services.AuthAccount;
+import ch.epfl.sdp.peakar.user.services.AuthService;
 import ch.epfl.sdp.peakar.utils.MenuBarTestHelper;
 
 import static androidx.test.espresso.Espresso.onView;
 import static androidx.test.espresso.assertion.ViewAssertions.matches;
+import static ch.epfl.sdp.peakar.user.AuthAccountTest.registerAuthUser;
+import static ch.epfl.sdp.peakar.user.AuthAccountTest.removeAuthUser;
+import static ch.epfl.sdp.peakar.utils.TestingConstants.BASIC_USERNAME;
+import static ch.epfl.sdp.peakar.utils.TestingConstants.LONG_SLEEP_TIME;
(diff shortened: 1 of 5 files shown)


app/src/main/java/ch/epfl/sdp/peakar/social/SocialListAdapter.java
```diff
@@ -111,4 +115,17 @@ private void setMedal(ImageView v, int rank) {
...
+    /**
+     * Set correct profile avatar
+     * @param v avatar view.
+     * @param avatarUrl avatar url of the social item.
+     */
+    private void setAvatar(ImageView v, Uri avatarUrl) {
+        if(avatarUrl == Uri.EMPTY) return;
+        Glide.with(mContext)
+                .load(avatarUrl)
+                .circleCrop()
```
aolsson711, line 128: I know that with the gallery it is kind of jacked since there is a lot that is done in the getItem function from the listAdapter. It is called alot of times when scrolling etc. 
I don't know a lot about Glide if it is fast or slow. But would it be possible to do some of this on the social item instead, like everything before `.into(v)` when we create the item? 

So we don't have to do the load, crop etc. as much?

aolsson711 (commented):
Nice work. I really like the global list test, with the update. 
The biggest comment I added was regarding if it is possible to move parts of the profile picture to the social item.

app/src/androidTest/java/ch/epfl/sdp/peakar/social/SocialActivityTest.java
```diff
@@ -25,9 +45,55 @@
...
+                // Check if list before sorting and list after sorting are equal
+                return socialItems.equals(sortedSocialItems);
+            }
+
+            @Override
+            public void describeTo(Description description) {
+                description.appendText("Users are sorted by their score");
+            }
+        };
+    }
```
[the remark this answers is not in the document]
giommok: Yeah, I also thought about it. If in the helper class we have also the matchers that are used in just one class (like in this case), maybe I should add also this one.

app/src/androidTest/java/ch/epfl/sdp/peakar/social/SocialActivityTest.java
```diff
@@ -71,6 +137,27 @@ public void TestTopBarColor() {
         onView(ViewMatchers.withId(R.id.top_bar)).check(matches(withBackgroundColor(R.color.LightGrey)));
     }
 
+    /* Test that the items in the global rankings are sorted correctly */
+    @Test
+    public void TestSortingInGlobalRankings() {
+        try {
+            Thread.sleep(LONG_SLEEP_TIME);
```
magangloff, line 144: Is there a reason why a sleep was added at the beginning of the test?

app/src/main/java/ch/epfl/sdp/peakar/social/SocialListAdapter.java
magangloff: I also have no experience with glide but it seemed really smooth to me on the phone and emulator.

magangloff (commented):
Looks really nice with the avatars and is also really smooth.
Added a question regarding a test.

app/src/main/java/ch/epfl/sdp/peakar/social/SocialListAdapter.java
giommok: As @magangloff pointed out by trying on the phone, it should be smooth even if these lines were called many times.
In fact, when `Glide` downloads an image, it is stored in a cache. This way, even if the method is called multiple times the image is downloaded only once (as you can read [here](https://stackoverflow.com/questions/41758087/glide-not-updating-image-android-of-same-url)).
The other operations on the downloaded file should be very fast and I think we can easily ignore them.

app/src/androidTest/java/ch/epfl/sdp/peakar/social/SocialActivityTest.java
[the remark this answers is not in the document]
aolsson711: Keep it here then

app/src/androidTest/java/ch/epfl/sdp/peakar/social/SocialActivityTest.java
giommok: Yes. As the rankings loading is not an immediate operation, it is necessary to wait for it.

aolsson711 (approved):
Nice work!

magangloff (approved):

PR #249, opened 2021-05-18
Fix bug language

magangloff:
Fixed bug #194 
Created a Utility class for the settings. Contains only helper methods to set the language for now.
The language will now be persistent even when the app is closed. A problem stays, when changing the language the already created activities stays in the old language. The only solution for this is to recreate them or close the application. The problem however is that when recreating activities will display them thus changing the activity. However, this should be not a problem anymore when the MainMenuActivity gets removed.
Also added new arrays.xml and string.xml files for the translations. I will update the german translations once we have the final UI and removed the unused strings.


app/src/main/java/ch/epfl/sdp/peakar/utils/SettingsUtilities.java
```diff
@@ -0,0 +1,77 @@
...
+        SettingsUtilities.setLocale(context, SettingsUtilities.getLanguageCode(language));
+    }
+
+    /**
+     * Get language code from language
+     *
+     * @param language language
+     * @return language code (2 letters)
+     */
+    private static String getLanguageCode(String language){
```
aolsson711, line 38: I am not sure which is best. Just to raise the question. But what is more scaleable to have this switch statement or would it be possible to have some kind of static final HashMap/Dictionary? What do you think?

app/src/main/java/ch/epfl/sdp/peakar/utils/SettingsUtilities.java
```diff
@@ -0,0 +1,77 @@
...
+        Configuration conf = res.getConfiguration();
+        conf.setLocale(myLocale);
+        //update locale configuration with new language
+        res.updateConfiguration(conf, dm);
+        Activity activity = (Activity) context;
+        activity.finish();
+        //Override transition and restart activity
+        activity.overridePendingTransition(0, 0);
+        activity.startActivity(activity.getIntent());
+        activity.overridePendingTransition(0, 0);
```
aolsson711, line 75: After we override the transition. Do we need to set it back or is that fine?

app/src/main/res/values-de-rDE/strings.xml
```diff
@@ -0,0 +1,185 @@
+<resources>
+    <string name="app_name" translatable="false">PeakAR</string>
```
aolsson711, line 2: Cirrus complains that this is translated in German. Since it is set to false it shouldn't

app/src/main/res/values-de-rDE/strings.xml
```diff
@@ -0,0 +1,185 @@
...
+
+    <!-- Google -->
+    <!-- Sign-in status messages -->
+    <string name="mountain_info" translatable="false">Mont Blanc - Monte Bianco lat: 45.8325 lon: 6.8641666666667 alt: 4810 <br /><br />
```
aolsson711, line 6: Cirrus complains that this is translated in German. Since it is set to false it shouldn't

app/src/main/res/values-de-rDE/strings.xml
```diff
@@ -0,0 +1,185 @@
...
+                                 Dent du Geant lat:45.86355980599387 lon:6.951348205683087 alt: 4013<br /><br />
+                                 Aiguille du plan lat:45.891667 lon:6.907222 alt: 3673<br /><br />
+                                 Pointe de Lapaz lat:45.920774986207014 lon:6.812914656881065 alt: 3660</string>
+    <!-- Sign-in button labels and other strings -->
+
+    <string name="title_activity_main_menu" translatable="false">MainMenuActivity</string>
```
aolsson711, line 12: Cirrus complains that this is translated in German. Since it is set to false it shouldn't

app/src/main/res/values-de-rDE/strings.xml
```diff
@@ -0,0 +1,185 @@
...
+
+    <!-- USERNAME CHOICE -->
+    <string name='already_existing_username'>The username is already used by another user. Choose a new one.</string>
+    <string name='invalid_username'>The username is not valid. It must contain between 3 and 15 alphanumeric characters.</string>
+    <string name='current_username'>You can\'t choose the username you already have!</string>
+    <string name='available_username'>Your username has changed!</string>
+    <string name='insert_username_button'>Insert your username here</string>
+    <string name='change_username_button'>Change username</string>
+    <string name='submit_username_button'>Change</string>
+    <string name='registered_username' translatable="false">Welcome to PeakAR community!</string>
```
aolsson711, line 32: This one should maybe actually be set to translatable!

app/src/main/res/values-de-rDE/strings.xml
```diff
@@ -0,0 +1,185 @@
...
+
+    <!-- FRIENDS SYSTEM -->
+    <string name='add_friend_button'>Add a friend</string>
+    <string name='friends_button'>Friends</string>
+    <string name='insert_friend_button' translatable="false">Insert here</string>
```
aolsson711, line 37: Cirrus complains that this is translated in German. Since it is set to false it shouldn't

app/src/main/res/values-de-rDE/strings.xml
```diff
@@ -0,0 +1,185 @@
+<resources>
```
aolsson711, line 1: In this file a lot of the strings have been translated but the boolean that if they should be translated is set to false. Cirrus does not like that!

aolsson711 (commented):
Good work! Really nice utility. 
I asked two questions regarding the implementation. 
Then in the German translation, quite a few of the strings that have "translatable=false" have been translated (or maybe just copied) and Cirrus is not happy!

app/src/main/res/values-de-rDE/strings.xml
magangloff: Oh yes, I just copied the default strings.xml (english version) and didn't remove the translatable=false strings

app/src/main/java/ch/epfl/sdp/peakar/utils/SettingsUtilities.java
magangloff: No, we do not have to set them back. This was added by @bgabriel998 to make the black screen disappear when reloading the application.

app/src/main/java/ch/epfl/sdp/peakar/utils/SettingsUtilities.java
magangloff: I think that in our case there are not so many translations so just a simple switch is probably enough.

app/src/androidTest/java/ch/epfl/sdp/peakar/general/InitActivityTest.java
```diff
@@ -32,10 +32,11 @@ public void cleanUp() {
 
     /* Test that the MainMenuActivity gets called */
     @Test
-    public void TestMainMenuActivityIsCalled() {
+    public void TestMainMenuActivityIsCalled() throws InterruptedException {
         //Launch InitActivity
         Intent intent = new Intent(ApplicationProvider.getApplicationContext(), InitActivity.class);
         ActivityScenario.launch(intent);
+        Thread.sleep(10);
```
ravifrancesco, line 39: Not that useful, but I would suggest using the TestingConstants class

app/src/main/java/ch/epfl/sdp/peakar/utils/SettingsUtilities.java
```diff
@@ -0,0 +1,77 @@
...
+        switch (language){
+            case "german":
+                return "de";
+            case "italian":
+                return "it";
+            case "french":
+                return "fr";
+            case "swedish":
+                return "sv";
+            case "english":
```
ravifrancesco, line 48: Why isn't there any return here?

ravifrancesco (commented):
Good job! Left a couple of comments

app/src/main/java/ch/epfl/sdp/peakar/utils/SettingsUtilities.java
magangloff: case "english" and and the default return the same value so I made the case "english" fall through. Basically, if the case is "english" then the flow of control will fall to the next case which is the default case that returns "en". I could also simply delete the case "english" but I think that it's better to display the case.

aolsson711 (approved):
Good work!

ravifrancesco (approved):

PR #250, opened 2021-05-20
Changed behaviour of Bresenham allgorithm

ravifrancesco:
Fixed a bug that caused the Bresenham algorithm to create empty lines of points in some cases.

Now the line of sight generates the correct line every for every POIPoint.

**TODO**
Fix other bugs that cause the line of sight to not work properly

### DIFF
--- a/app/src/main/java/ch/epfl/sdp/peakar/points/LineOfSight.java
+++ b/app/src/main/java/ch/epfl/sdp/peakar/points/LineOfSight.java
@@ -137,12 +137,13 @@ public class LineOfSight {
      */
     private List<Pair<Integer, Integer>> bresenham(int x1, int y1, int x2, int y2) {
 
-        int m_new = 2 * (y2 - y1);
-        int slope_error_new = m_new - (x2 - x1);
+        int m_new = 2 * Math.abs(y2 - y1);
+
+        int slope_error_new = m_new - Math.abs(x2 - x1);
 
         List<Pair<Integer, Integer>> line = new ArrayList<>();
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/ch/epfl/sdp/peakar/points/LineOfSight.java
```diff
@@ -152,11 +153,13 @@ private boolean isVisible(POIPoint poiPoint, Pair<Integer, Integer> userIndexes,
...
             if (slope_error_new >= 0) {
-                y++;
-                slope_error_new -= 2 * (x2 - x1);
+                y = y2 > y1 ? y + 1 : y - 1;
+                slope_error_new -= 2 * Math.abs(x2 - x1);
             }
         }
 
+
+
```
giommok, line 162: Empty lines
```suggestion
```

app/src/main/java/ch/epfl/sdp/peakar/points/LineOfSight.java
```diff
@@ -152,11 +153,13 @@ private boolean isVisible(POIPoint poiPoint, Pair<Integer, Integer> userIndexes,
             // Slope error reached limit, time to
             // increment y and update slope error.
```
giommok, line 154: I think the comment was describing the old algorithm.
```suggestion
            // increment or decrement y and update slope error.
```

giommok (changes requested):
Good job at finding that bug!
I left two comments.

bgabriel998 (approved):
Not much to say 👍

giommok (approved):

Issue #110, opened 2021-04-11
Bug report: Mountain marker and compass UI does not stay at same position


magangloff:
## Expected Behavior

The mountain markers and compass pointing to a position on the camera-preview should not point to another position when rotating the device.

## Current Behavior

When rotating the device horizontally, the compass points in another direction. The same goes for the mountain markers
When rotating the device vertically, the mountain markers do not stay at the same position on the camera preview but move with the device rotation: When rotating upwards, the mountain markers go up instead of resting at the same place. When rotating downwards, the mountain markers go down instead of resting at the same place. 

## Steps to Reproduce

1. Open camera-preview
2. Wait until the mountain markers are displayed
2. Rotate the device horizontally and vertically

PR #259, opened 2021-05-21
Fix bug multiple download

magangloff:
Could not replicate bug #234 
The bug has maybe been solved or updates of location were wrong when the bug occurred.
Addressed instead bug #110  where the markers and the compass on the camera-view would move and not stay at the same location.
The problem was, that the calculated field of view was the field of view for a 4:3 aspect ratio. However since we are using a fullscreen aspect ratio, the field of view needs to be adapted by the aspect ratio.
Also changed some minor UI for the dev options on the camera preview.
Addressed bug #237 where there would be an offset between displaying the markers in portrait or landscape left/right. 
The problem was, that the coordinate system was not remapped correctly depending on the orientation of the device.


app/src/main/java/ch/epfl/sdp/peakar/camera/Compass.java
```diff
@@ -100,6 +110,37 @@ public void onSensorChanged(SensorEvent event) {
...
+                axisZ = SensorManager.AXIS_MINUS_Z;
+                break;
+
+            case Surface.ROTATION_270:
+                axisX = SensorManager.AXIS_MINUS_Z;
+                axisZ = SensorManager.AXIS_X;
+                break;
+
+            default:
+                break;
```
bgabriel998, line 140: Either put `default:` on top of another case or remove it completely.

bgabriel998 (commented):
Nice,  juste take the comment on the switch case in consideration before approval

giommok (approved):
Good job, nothing to say!

bgabriel998 (approved):

PR #268, opened 2021-05-22
New UI

### LINKED ISSUE #254: #197 Connect NewUI with App
The new UI is still activities on the side. Implement so that the whole app is using the correct UI instead.

aolsson711:
implemented new UI. Old activities exist but are not reachable from application.
close #254


bgabriel998 (approved):
Nice refactor ! Approved

bgabriel998:
What appened with the `ToolbarHandler.java` file?  Code coverage is F

magangloff (changes requested):
Looks really good. If I'm correct, every time you select an activity through the menu bar, a new intent is created. We should however check if there is already an active activity to not create a new one every time cause this would make the app really slow after some time.
You can check out this page for more information: https://developer.android.com/guide/components/activities/tasks-and-back-stack

Another question that I have is if the user presses the back-button, should he be brought back to the last activity that he was at our should he be brought back to a certain activity like the camera for example?

magangloff (changes requested):
Found a solution for the handling of the activity creation.

magangloff (approved):

PR #269, opened 2021-05-22
SocialActivity and ProfileActivity updates

giommok:
# SocialActivity #257 

I've modified the old `RemoteFriendItem` to extend `SocialItem`.
The new methods are to set and remove the listener for changes in the friend on the DB.

Then, I've added a method called `synchronizeFriends`, similar to the old `synchronizeGlobal`, and a method `produceFriendItemListener` that will be used by `synchronizeFriends` to set a new listener to the `RemoteFriendItem`.

Furthermore, I've modified the `SocialListAdapter` so that it is possible to filter the results with the `EditText` in `SocialActivity`.

Finally, now it is possible to click on social items. After the click, the profile of the user is opened. If authenticated users click on their own social items, they will be redirected to their authenticated profile. If a user is not authenticated, on the profile all the "friends" buttons will be hidden.

# (New)ProfileActivity #256 

First, I've removed the old profile classes. I've not renamed the new profile activity because I know some of us are working on the same class and I don't want to cause any problems with possible conflicts. 

Then, I've connected the `OtherAccount` with the shown data. Finally, I've implemented the add/remove friend features and added tests for these functionalities. Moreover, for other users, it is possible to refresh their profile with a gesture.

# Database #258

I've added two methods to set online/offline mode. I also tested these methods, so they do work properly.

## How does it work?

Directly from Firebase documentation:

> The Firebase Realtime Database stores data returned from a query for use when offline. For queries constructed while offline, the Firebase Realtime Database continues to work for previously loaded data. If the requested data hasn't loaded, the Firebase Realtime Database loads data from the local cache. When network connectivity is available again, the data loads and will reflect the query.
[...]
Any transactions that are performed while the app is offline, are queued. Once the app regains network connectivity, the transactions are sent to the Realtime Database server.

For more details, [this](https://firebase.google.com/docs/database/android/offline-capabilities?hl=en) is the link to the quote.


app/src/main/java/ch/epfl/sdp/peakar/database/Database.java
```diff
@@ -54,4 +61,20 @@ public static Database getInstance() {
...
     }
+
+    /**
+     * Enable the DB offline mode.
+     * Requests for which data is cached can still be answered.
+     * Other requests will be enqueued until online mode is set.
+     */
+    public void setOfflineMode() {
+        FirebaseDatabase.getInstance(DATABASE_ADDRESS).goOffline();
+    }
```
ravifrancesco, line 72: I have not understood how this is called, shouldn't there be a `SharedPreferenceListener`?

app/src/main/java/ch/epfl/sdp/peakar/social/RemoteFriendItem.java
```diff
@@ -0,0 +1,37 @@
+package ch.epfl.sdp.peakar.social;
+
+import com.google.firebase.database.ValueEventListener;
+
+import ch.epfl.sdp.peakar.database.Database;
+import ch.epfl.sdp.peakar.database.DatabaseReference;
+import ch.epfl.sdp.peakar.social.SocialItem;
```
ravifrancesco, line 7:
```suggestion
```

app/src/main/java/ch/epfl/sdp/peakar/user/services/AuthAccount.java
```diff
@@ -6,8 +6,8 @@
 
 import ch.epfl.sdp.peakar.points.CountryHighPoint;
 import ch.epfl.sdp.peakar.points.POIPoint;
+import ch.epfl.sdp.peakar.social.SocialItem;
 import ch.epfl.sdp.peakar.user.challenge.Challenge;
-import ch.epfl.sdp.peakar.user.friends.FriendItem;
 import ch.epfl.sdp.peakar.user.outcome.ProfileOutcome;
```
ravifrancesco, line 12: Method `isEnrolledInAChallenge()` is never used

ravifrancesco (changes requested):
Good job, the code looks good! I left some comments

app/src/main/java/ch/epfl/sdp/peakar/database/Database.java
giommok: Indeed it is not called, I've just implemented the method to be called in order to disconnect the DB.

app/src/main/java/ch/epfl/sdp/peakar/database/Database.java
```diff
@@ -1,7 +1,11 @@
 package ch.epfl.sdp.peakar.database;
 
+import com.google.firebase.database.FirebaseDatabase;
```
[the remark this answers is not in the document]
giommok: I don't think it would be wise to remove those lines for now, because Bastien is working on the challenges part and maybe is using them.

app/src/main/java/ch/epfl/sdp/peakar/database/Database.java
```diff
@@ -44,7 +48,10 @@ private Database() {
      * Get a database instance.
      */
     public static Database getInstance() {
-        if(instance == null) instance = new Database();
+        if(instance == null) {
+            instance = new Database();
+            FirebaseDatabase.getInstance().setPersistenceEnabled(true);
+        }
         return instance;
```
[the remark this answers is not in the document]
giommok: I will do it!

app/src/main/java/ch/epfl/sdp/peakar/user/services/AuthAccount.java
giommok: As before, Bastien is working on the challenges part so I don't think we should remove anything related to it for now.

app/src/main/java/ch/epfl/sdp/peakar/social/RemoteFriendItem.java
```diff
@@ -0,0 +1,37 @@
...
+
+public class RemoteFriendItem extends SocialItem {
```
aolsson711, line 8: Missing java doc

app/src/main/java/ch/epfl/sdp/peakar/user/profile/NewProfileActivity.java
```diff
@@ -305,11 +329,89 @@ public boolean dispatchTouchEvent(MotionEvent ev) {
...
+                });
+            }
+        };
+        removeThread.start();
+    }
+
+    /**
+     * On social activity button click
+     */
+    public void socialActivityButton(View view) {
```
aolsson711, line 413: What is this?

aolsson711 (commented):
Great work! Very small comment, on missing javadoc for class. And also a question why there is a social button function in the profile activity?

app/src/main/java/ch/epfl/sdp/peakar/user/profile/NewProfileActivity.java
giommok: It is the "friends" buttons, that actually loads the `SocialActivity`.

app/src/main/java/ch/epfl/sdp/peakar/social/RemoteFriendItem.java
giommok: Added.

app/src/main/java/ch/epfl/sdp/peakar/user/profile/NewProfileActivity.java
aolsson711: You mean the icon in the top right corner? I don't think the idea was that it was supposed to be a button. More an icon with different color if you are friends or not

giommok: Oh, I thought this was a button for the `SocialActivity`. Anyway, you can see if you are friends because you have the add/remove button so it would be redundant. Personally, I like it this way. What do you think?

ravifrancesco (approved):
Good job!

aolsson711 (approved):
Great work!

PR #272, opened 2021-05-26
Country icons

### LINKED ISSUE #255: #197 Add correct country Icon to collected
Currently the flag is hard-coded to France. Change so it is the correct Country displaying that.

aolsson711:
Added so that the country flag of a collected is set correctly

closes #255


magangloff (changes requested):
Would be nice to refactor the countries flag into a separate folder under drawable to get a better overview of the existing files. Here is a link on how to do that. Basically, you simply create a new folder under /drawable like /drawable/countries and then in that folder you need to add a drawable folder with all the flags. At the end, you would have something like /res/drawable/countries/drawable.
https://proandroiddev.com/split-layout-into-subfolder-in-android-5bf76f805e90

app/src/main/java/ch/epfl/sdp/peakar/collection/NewCollectedItem.java
```diff
@@ -31,14 +32,15 @@
      * @param date item got collected.
      */
     public NewCollectedItem(String name,int points, int height, boolean topInCountry,
-                            float longitude, float latitude, String date) {
+                            float longitude, float latitude, String date, String country) {
```
magangloff, line 35: Missing javadoc for the country param

app/src/main/java/ch/epfl/sdp/peakar/collection/NewCollectionListAdapter.java
```diff
@@ -1,12 +1,17 @@
 package ch.epfl.sdp.peakar.collection;
 
 import android.content.Context;
+import android.location.Address;
+import android.location.Geocoder;
+import android.util.Log;
```
magangloff, line 6: Not used imports can be removed.
```suggestion

```

app/src/main/java/ch/epfl/sdp/peakar/collection/NewCollectionListAdapter.java
```diff
@@ -1,12 +1,17 @@
...
 import android.view.View;
 import android.view.ViewGroup;
 import android.widget.ArrayAdapter;
+import android.widget.ImageView;
 import android.widget.TextView;
 
 import java.util.List;
+import java.util.Locale;
```
magangloff, line 14: Not used import can be removed.
```suggestion

```

app/src/main/java/ch/epfl/sdp/peakar/user/profile/NewProfileActivity.java
```diff
@@ -439,4 +445,24 @@ private void showErrorMessage() {
...
         snackbar.show();
     }
+
+    /**
+     * Retrieve country using input latitude and longitude
+     * @param latitude poiPoint latitude (peak latitude)
+     * @param longitude poiPoint longitude (peak longitude)
+     * @return Name of the country where the peak is located
+     */
+    private String getCountryFromCoordinates(double latitude,double longitude){
```
magangloff, line 455: Should this method be moved in a utility class?

magangloff (changes requested):
Nice, this makes the drawable folder a lot cleaner. I think that we will be able to also do this once we have removed the drawables that are not used anymore.

bgabriel998 (approved):
Nothing to say, well done. 

_Note that the function that return the country from coordinates returns only one country if the peak is located on a border (which is very likely to happen since the borders are based on mountains)_

magangloff (approved):

PR #273, opened 2021-05-27
Fixing line of sight bugs

ravifrancesco:
Fixed the line of sight bugs:

- The line computed between the points doesn't have any gap now
- The filtering seems to work correctly now. I have used [Siolwise Elevation Tool](https://www.solwise.co.uk/wireless-elevationtool.html) to check if the results are correct

The Bresenham algorithm has been changed as it had issues with some points, the one now works fine, except for a bug with points exactly on the vertical axis.

The Bounding Box range from the settings is now used across all the app.


magangloff (approved):

bgabriel998 (approved):
Nice one !

PR #282, opened 2021-06-01
Fixing last bugs line of sight

ravifrancesco:
Now the Line of Sight has no more bugs!

Things that were changed:
- Fixed the way the ratio is computed in case of a difference in indexes equal to 0 in the `drawLine` method
- Fixed test `getVisiblePointsTest` (one of the point was tested as nonvisible while being visible in real life)
- Fixed test `getVisiblePointsLabeledTest`
- Fixed test `getAltitudeAndIndexesTest`
- Removed TODOs that were solved

The app now won't save `null` offline content. If the download fails, it will return to the `SettingsActivity`, reset the `offline_mode_key` to `false` and display a toast


app/src/androidTest/java/ch/epfl/sdp/peakar/points/TopographyAsyncTest.java
```diff
@@ -101,166 +102,127 @@ public void getBoundingBoxWestLongTest() {
...
+        Assert.assertNotNull(indexes.first);
+        Assert.assertNotNull(indexes.second);
+        Assert.assertEquals(8849, elevationMap.getAltitudeAtLocation(indexes.first, indexes.second), 200);
 
-        Map<POIPoint, Boolean> filtered = lineOfSight.getVisiblePointsLabeled(pointsToCheck);
+        int[][] oldTopographyMap = topographyPair.first;
 
-        Log.d("DEBUG", filtered.toString());
+        // set location near the Mont Blanc
+        userPoint.setLocation(45.802537, 6.850328, 0, 0);
```
bgabriel998, line 206: Location could be added to the Testing constants

app/src/androidTest/java/ch/epfl/sdp/peakar/points/TopographyAsyncTest.java
```diff
@@ -101,166 +102,127 @@ public void getBoundingBoxWestLongTest() {
...
+        // check the altitude around the Everest Peak using coordinates
+        Assert.assertEquals(8849, elevationMap.getAltitudeAtLocation(GPSTracker.DEFAULT_LAT, GPSTracker.DEFAULT_LON), 200);
+        // check the altitude around the Everest Peak using indexes
+        Pair<Integer, Integer> indexes = elevationMap.getIndexesFromCoordinates(GPSTracker.DEFAULT_LAT, GPSTracker.DEFAULT_LON);
+        // check the indexes around the Everest Peak
+        Assert.assertEquals(new Pair<>(539, 610), indexes);
+        Assert.assertNotNull(indexes);
+        Assert.assertNotNull(indexes.first);
+        Assert.assertNotNull(indexes.second);
+        Assert.assertEquals(8849, elevationMap.getAltitudeAtLocation(indexes.first, indexes.second), 200);
```
bgabriel998, line 201: Elevation could be added to testing constants

app/src/androidTest/java/ch/epfl/sdp/peakar/points/TopographyAsyncTest.java
```diff
@@ -101,166 +102,127 @@ public void getBoundingBoxWestLongTest() {
...
+        // check the altitude around the Mont Blanc Peak using indexes
+        indexes = elevationMap.getIndexesFromCoordinates(45.8326, 6.8652);
+        // check the indexes around the Mont Blanc Peak
+        Assert.assertEquals(new Pair<>(502, 790), indexes);
+        Assert.assertNotNull(indexes);
+        Assert.assertNotNull(indexes.first);
+        Assert.assertNotNull(indexes.second);
+        Assert.assertEquals(4808, elevationMap.getAltitudeAtLocation(indexes.first, indexes.second), 200);
+        // check the altitude around the Mont Blanc Peak using coordinates
+        Assert.assertEquals(4808, elevationMap.getAltitudeAtLocation(45.8326, 6.8652), 200);
```
bgabriel998, line 225: Same comment about constants

app/src/main/java/ch/epfl/sdp/peakar/points/LineOfSight.java
```diff
@@ -167,9 +162,9 @@ private boolean isVisible(POIPoint poiPoint, Pair<Integer, Integer> userIndexes,
         double cumulatedError = 0;
 
         if (Math.abs(pixelY) >= Math.abs(pixelX)) {
-            ratio = pixelX != 0 ? Math.abs(pixelY / pixelX) : Math.abs(pixelX);
```
bgabriel998, line 170: Check code climate issue : similar code blocks.
Couldn't we define a method instead ?

bgabriel998 (commented):
Consider comments

app/src/androidTest/java/ch/epfl/sdp/peakar/points/TopographyAsyncTest.java
```diff
@@ -101,166 +102,127 @@ public void getBoundingBoxWestLongTest() {
...
-        Log.d("DEBUG", filtered.toString());
+        // set location near the Mont Blanc
+        userPoint.setLocation(45.802537, 6.850328, 0, 0);
+        elevationMap.updateElevationMatrix();
 
-        Assert.assertTrue(true);
+        //Wait for the map to be updated
+        int counter=0;
+        while(elevationMap.getTopographyMap()==oldTopographyMap && counter<50){
+            Thread.sleep(1000);
```
giommok, line 212: In the testing constants class, we have a constant called `LONG_SLEEP_TIME` (or something like this) that you could use instead of a hard-coded constant.

giommok (changes requested):
Good job. 
I agree with Bastien's suggestions, I also added one more comment about a constant that we already have and you can use.

app/src/androidTest/java/ch/epfl/sdp/peakar/points/TopographyAsyncTest.java
ravifrancesco: I actually didn't code this part of the test. @magangloff do you agree with @giommok suggestion?

app/src/main/java/ch/epfl/sdp/peakar/points/LineOfSight.java
ravifrancesco: I could put these similar blocks in one method, but this method would require this parameters:
* `x1`
* `x2`
* `y1`
* `y2`
* `ratio`
* `error`
* a `boolean` to indicate if the line is computed by circling on the x or y axis

I don't know it it is a good idea to have a method with this many parameters that would be 7-8 lines long

bgabriel998 (approved):

app/src/androidTest/java/ch/epfl/sdp/peakar/points/TopographyAsyncTest.java
magangloff: Yes, you can simply use THREAD_SLEEP_1S or LONG_SLEEP_TIME from the constants, the hard-coded constant was before we introduced the testing constants

giommok (approved):

PR #284, opened 2021-06-02
User interaction improvements

giommok:
# User interaction
- Improved the user interface with loading dialogues while a profile is being loaded.
- Improved the app response when no connection is available. Now it is way more robust. In addition, the profile of the user is no more available for a user that is offline (the usual error message is shown). But one could say that by pressing the back button you would still be able to go to the profile if it was already loaded before the connection was down. Yes, it's possible indeed. But in this case, error messages would be shown upon pressing the buttons to add friends/remove friends/change username, exactly as before.

# Backend

- Improved how the account was loaded when starting the app. Now, a check of the connection is done first. If the connection is available, the account is loaded. Otherwise, it is not (and, in case the connection returns, the auth account data will be retrieved after pressing the profile button).
- The account data retrieval in the `InitActivity` is **not asynchronous** anymore. Thus, if the user was already authenticated and has an internet connection, the account data will be retrieved before the `InitActivity` is dispatched.

# Tests
Improved some tests and moved the widely used mock user creation methods to a separate utility class (as, before, they were located in the `AuthAccountTest` class).


bgabriel998 (approved):
Approved, nothing to say here 👍

app/src/main/java/ch/epfl/sdp/peakar/general/InitActivity.java
```diff
@@ -38,12 +40,22 @@
...
 
     private MultiplePermissionsListener allPermissionsListener;
+    private Context mContext;
 
     @Override
     protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
+
+        mContext = this;
+
```
magangloff, line 50: mContext is not used and can be removed.
```suggestion

    @Override
    protected void onCreate(Bundle savedInstanceState) {
        super.onCreate(savedInstanceState);
```

app/src/main/java/ch/epfl/sdp/peakar/user/profile/ProfileLauncherActivity.java
```diff
@@ -37,45 +50,68 @@ protected void onCreate(Bundle savedInstanceState) {
...
-        authService = AuthService.getInstance();
+        if(!Database.getInstance().isOnline()) {
+            AlertDialog.Builder builder = new AlertDialog.Builder(this);
+            builder.setMessage(ProfileOutcome.FAIL.getMessage())
+                    .setCancelable(false)
+                    .setPositiveButton("Ok", new DialogInterface.OnClickListener() {
+                        public void onClick(DialogInterface dialog, int id) {
+                            finish();
+                        }
+                    });
```
magangloff, line 61: Can be replaced by lambda expression.
```suggestion
                    .setPositiveButton("Ok", (dialog, id) -> finish());
```

magangloff (changes requested):
Nice improvements, added just some small suggestions for code readability.

magangloff (approved):

